PR #8, opened 2015-02-10
Getting rid of the "concepts" chapter

cconcolato:


heycam:
The rest looks good.

cconcolato:
"s/vector graphics/documents/" doesn't look right to me, for instance with "mixed with HTML content". "vector graphics" is more generic.

The change for namespaces is ok.

Your third proposed change does not carry the fact script can be used to do any manipulation, not necessarily animations.

heycam:
OK for the first comment.  But I think "SVG vector graphics" sounds like "ATM machine". :)

For the third change, then, what about something like this:

SVG also supports dynamic changes; script can be used to create interactive documents, and animations can be performed using declarative animation features or by using script.

PR #13, opened 2015-03-04
`a` element target attribute

AmeliaBR:
Removed the `_replace` keyword, added note explaining why, and updated
other definitions to harmonize with (and reference to) HTML 5
definitions of browsing contexts.

I left in an issue note, as there remain a few differences between SVG and HTML (e.g., case sensitivity) that I think are reasonable but others may disagree.


Nantonos:
Looks good to me after reviewing the changes. Browsing context did not exist when the original SVG 1.x text was written, it would have made things easier to explain.

PR #28, opened 2015-12-02
Fix a typo in the SVGViewElement IDL.

Ms2ger:


nikosandronikos:
Agree. The grammar requires the braces be present.
Thanks!

PR #29, opened 2015-12-03
Take the excludefromidl attribute into account.

Ms2ger:
During the generation of the spec pages that contain the IDL, the attribute
was being removed, causing the index generation code to never see an element
with the attribute present.


heycam:
Thanks for looking at this!

Issue #102, opened 2016-04-13
Clean up language re <title>/<desc> and tooltips


nikosandronikos:
https://svgwg.org/svg2-draft/struct.html#issue20

I don't think there's anything important in Tiny 1.2 that isn't in SVG 2. Particularly, with the resolution of issue #100.
I'll open the issue in case anyone else has differing views and we can review in London.

AmeliaBR:
(and clean-up re SVG accessibility issues for empty title/desc and multilingual)

nikosandronikos:
Removed issue from the spec relating to taking text from SVG Tiny 1.2 in commit 41fa130cf61356727eb33de6c8b483d1ccaf9a1f

AmeliaBR:
Also need to clean up re which elements can have a title/desc child.  Currently the prose says "container" and "graphics" elements can have title & desc, but I think other elements allow descriptive elements in their content models, while `switch` elements now explicitly exclude them (see [discussion in SVG Accessibility mailing list](https://lists.w3.org/Archives/Public/public-svg-a11y/2015May/0057.html)).

nikosandronikos:
This is editorial, but needs to be rewritten in sync with the accessibility specs.

AmeliaBR:
Cleaning up which elements can have title/desc extends to the FXTF specs, see https://github.com/w3c/fxtf-drafts/issues/9

jarek-foksa:
Besides container and graphics elements, are you planning to allow descriptive elements as children of anything else?

AmeliaBR:
My instinct is to allow them generously, except for where there is a direct logical conflict (like `switch`).  We would need to be clear that a non-rendered element is not exposed to assistive tech simply because it has a description.  However, you could use the description of the graphical effect element to generate a full description of the graphics:

```
<pattern id="dots"... >
  <title>blue and white polka dot pattern</title>
  ...
</pattern>

<path d="..." fill="url(#dots)" aria-describedby="dots">
  <title>Bow tie</title>
</path>
```

The phrase "blue and white polka dot pattern" would be available as an extra description on the shape because of the `aria-describedby` attribute.  It could also be used simply as metadata in authoring tools to describe the pattern itself.  But there would not be a pattern object in the accessibility tree when you view the graphic in a browser.
   
The alternative approach would be to forbid `<title>` and `<desc>` on all never-rendered elements.  As @jarek-foksa notes, the current haphazard & inconsitent approach doesn't make sense.

[...]

PR #170, opened 2016-06-20
Title & desc re-write

### LINKED ISSUE #101: Define purpose of other namespaced markup in <title> and <desc>
https://svgwg.org/svg2-draft/struct.html#issue23

Prior discussion, from https://www.w3.org/2015/03/05-svg-minutes.html#item12:
> 
> ed: I think the issue is to say what this purpose is
> 
> heycam: presumably to include some more structured descriptions rather than just plain text
> 
> AmeliaBR: and originally also so that you could have complex fallback content, in a browser that didn't support SVG
> ... that's no longer relevant though
> ... I'm pretty sure the main purpose is so you can have structured content inside `<desc>`
> 
> heycam: note also the HTML parsing algorithm allows HTML content inside here
> 
> AmeliaBR: I'm not sure if there's much of a benefit within `<title>`, but some browsers do support `<br>` within `<title>` and some don't
> ... a weird random thing
> ... maybe that should be cleaned up
> 
> heycam: within SVG documents or HTML title?
> 
> <AmeliaBR> http://stackoverflow.com/questions/23626422/tooltip-to-display-in-two-separate-lines-not-working-in-firefox/25690802#25690802
> 
> AmeliaBR: so this may have been coming from this comment that external namespaces are OK
> ... so some support `<br>` within title
> 
> Tav: I don't really like the idea of having `<br>` in there
> ... maybe white-space:pre-line?
> 
> AmeliaBR: that's the other thing that's changed since SVG 1.1
> ... in the title="" in HTML you can put line breaks
> ... so either way isn't conforming to specs, but all browsers allow one way or the other
> ... either with `<br>` or a hard line break
> 
> heycam: and that's a hard line break without setting the white-space property?
> 
> AmeliaBR: yes
> ... and only for `<title>`
> 
> Tav: putting a `<br>` seems wrong to me
> ... allowing white space to be preserved seems more compatible
> 
> AmeliaBR: and with the new white-space option, I think that will be more reasonable
> 
> Tav: you can style your title
> 
> AmeliaBR: it's tricky, because tooltips aren't styled with CSS
> ... I have no idea whether white-space:pre would apply
> 
> heycam: do you think we need a issue in here to define this kind of thing?

AmeliaBR:
Many editorial changes, a few normative ones:

- Move the blue boxes down to be associated with the sections specific to title and desc, respectively (after general rules that apply to both)
- Re-arrange a number of sentences/paragraphs to group together normative text for authors vs user agents
- Clarify rules & add example for matching multi-lingual descriptive text
- SHOULD > MUST for user agent exposing text content to accessibility APIs
- Authors SHOULD NOT, authoring tools MUST NOT include empty title/desc (this causes a mess on the accessibility side)
- Advice for which elements should have title/desc
- Purpose of title/desc on non-rendered content
- Clean up the examples of purpose of title & desc
- Mention aria-labelledby when warning against redundant title content
- User agents SHOULD make titles available to all users on interaction (tooltips still being only one possible way of doing so)
- Clarify how this differs in practice for title on the root element
- Clean up paragraph on aria-describedby
- remove generic example (redundant given the earlier example)
- remove inaccurate statements about alternative plain-text presentations
- add warning that other namespace markup within title/desc has no affect on accessible alternative text (clears spec issue 23)

Addresses most of #102, fixes #101

### DIFF
--- a/master/struct.html
+++ b/master/struct.html
@@ -447,14 +447,6 @@ and <span class='element-name'>'title'</span> elements</h2>
   <edit:elementcategory name='descriptive'/>.</dd>
 </dl>
 
-<div id='DescElement'>
-<edit:elementsummary name='desc'/>
-</div>
-
-<div id='TitleElement'>
-<edit:elementsummary name='title'/>
-</div>
-
 <p class="note">The attribute <a>'lang'</a> added to allow internationalization
 of the <a>'desc'</a> and <a>'title element'</a> elements.</p>
 
--- a/master/struct.html
+++ b/master/struct.html
@@ -460,66 +460,117 @@ whose content is text. <a>'desc'</a> and <a>'title'</a> elements are
 not visually rendered as part of the graphics.</p>
 
 <p>Multiple sibling <a>'desc'</a> or <a>'title'</a> elements must have
-different <a>'lang'</a> attributes. The element whose <a>'lang'</a> attribute best
-matches the language set by the user agent will be used as the <a>'title'</a> or
-<a>'desc'</a>. If no match exists, the first element must be used. If multiple equally
-valid matches exist, the first match is used.</p>
+different languages, 
+as defined using a <a>'lang'</a> attribute (or <code>xml:lang</code> attribute) on the descriptive element or an ancestor. 
+The user agent must select the element of each type whose language best
+matches language preferences set by the user. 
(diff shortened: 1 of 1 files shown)


AmeliaBR:
The Editor's accepted these changes at the Amsterdam F2F, 21 June 2016.

PR #202, opened 2016-07-12
implnote.html: fix θ₁ → Δθ

waldyrious:


nikosandronikos:
Thanks!

waldyrious:
@nikosandronikos should I do anything to prevent further PRs from erroring in ash-nazg? I already created an account there by logging in with my github credentials.

nikosandronikos:
@waldyrious Nothing you need to do. It's just an indication that you're not a member of the SVG working group and so haven't made IP commitments. I just mark the PR non-substantive in the repo manager and then it goes green.

waldyrious:
Ok, thanks for the clarification. By the way, I did try to do that but the server seemed to be having some trouble updating the status. Might be something to be checked out.

Issue #121, opened 2016-04-21
Limit pattern & gradient attributes that are inheritable using href 


jarek-foksa:
The spec says that when you have a pattern element referencing another pattern element then:

> (...) Any attributes which are defined on the referenced element which are not defined on this element are inherited by this element.  ([source](https://svgwg.org/svg2-draft/pservers.html#PatternElementHrefAttribute))

Should it really state "any attributes"? It doesn't make sense to inherit e.g. `id`, `href` or `onload` attributes. I would limit the inheritable attributes to pattern-specific ones, namely:
- `x`
- `y`
- `width`
- `height`
- `viewBox`
- `preserveAspectRatio`
- `patternTransform`
- `patternUnits`
- `patternContentUnits`

AmeliaBR:
Good points.  And probably easier to list the attributes that _are_ copied instead of those that aren't.  As you say, event handlers shouldn't fire multiple times, and custom data attributes probably shouldn't be copied.

I'd also like clear language about how presentation attributes/styles inherit.  I'd assumed it would work like a <use> element, but no one seems to implement it that way.

nikosandronikos:
https://www.w3.org/2016/04/22-svg-minutes.html#resolution06

nikosandronikos:
I added the list above.
Wonder should we also inherit style since it might define width and height?

AmeliaBR:
It don't think it makes sense to copy `style` but not styles set using CSS rules.  I think we'll just have to define special rules for width & height, similar to how `<use>` has special rules for passing along width and height on a re-used symbol/svg.

AmeliaBR:
We don't currently treat any of the geometric attributes on paint servers as presentation attributes.  That avoids the issue with whether href cross-references should inherit computed style values.  

So the attribute inheritance can be done by a simple list of attributes, and checking for whether the current element has the attribute specified, no need to deal with cascade & used values.  I'm going to add similar lists of propagated attributes to all the other paint server types.

The same edits will synchronize the text with the new [URL & external resource processing rules](https://github.com/w3c/svgwg/pull/206/commits/efbe8d73885d235ed057327b5461a11fe6d3f7c3) and with shadow DOM structure for duplicated nodes.

Issue #44, opened 2016-02-05
Does conditional processing mute sound playback?


karip:
Does sound play for audio and video elements which are hidden by conditional processing?

The current text only mentions rendering, but it doesn't say how events, animations or sound are affected. An improved wording to address events and animations was suggested: “All others will not be rendered (as if the display property were set to none.)” [1]

However, that also suggests that sound is always played.

Can you add the improved wording and mention how sound is affected?

[1] https://lists.w3.org/Archives/Public/www-svg/2014Mar/0049.html

AmeliaBR:
To answer your question: yes, conditional processing would be expected to affect audio and video playback.  I am almost certain that `display: none` would also prevent audio/video from playing, but I couldn't find a good spec reference for that.

I do agree that the language could be improved to address audio/video.  In addition to whether resources are played, there is the question of whether user agents should download content that they aren't going to use.  I don't know what current behavior is for `<image>` in `<switch>` on that point.

tabatkins:
Yes, display:none stops `<audio>`/`<video>`.  I think HTML specifies this now?

AmeliaBR:
I can't find any spec in HTML of CSS that discusses the interaction of `display` with `<audio>` and `<video>`.  Quick testing shows that (in at least some browsers) `display` affects visual rendering of videos, but does not pause video progress or affect audio.

I still think it would be useful to have audio/video disabled by SVG conditional processing, but we'll need to be explicit: it should disable all processing, including file download.  Not sure how this disabled state should be reflected in the HTML DOM APIs, though. 

I've added an issue to the spec itself, as I cleaned up some of the related text.

karip:
After thinking about this, I would expect `<audio>` and `<video>` to behave like `<image>`. For instance, `<use>` referencing `<video>` under `<switch>` should show it playing. This means that `<video>` must be fetched and played back even if conditional processing hides it.

CSS Media queries have the same issue with audio and video. Audio and video get played although media queries have hidden them.

Since `<switch>` and media queries both have the same issue, I think that they should be solved in the same way. `<audio>` and `<video>` are HTML elements, so I would let HTML/CSS solve this issue (by not playing when display is none or some other means).

So, this isn’t really an SVG related issue and should be just closed.

AmeliaBR:
@karip 

[...]

Issue #184, opened 2016-07-03
Add an option to <use> elements to create independent, unlinked clones


AmeliaBR:
The copied graphics created by a `<use>` element are live mirrors of the original referenced DOM elements.  They must reflect any mutations on the original, and cannot be modified directly.

This means that the cloned elements cannot be fully animated or manipulated independently.  It also means that implementations must maintain mutation listeners on the source DOM fragments.  And it means that interactive content is a usability mess (see #183).

I know of multiple SVG authors who currently use JS frameworks to clone SVG document fragments because `<use>` elements just don't offer the necessary independence of each instance.

I am therefore proposing a new attribute on the `<use>` element, which would switch it from being a linked mirror of the referenced DOM fragment to being an independent clone.

Attribute name could be `method` or `shadow` (or `shadow-method`).  Values would be `clone` or `mirror`, with initial value `mirror`, to reflect the current behavior.

For static SVG renderers, this attribute would have no effect; it would only impact animation and scripting.  And probably only interactive animation at that, but I'm not certain without writing up a full proposal.

For script-supporting user agents, it would require very little additional complexity beyond the switch to a full shadow-DOM element model (as opposed to the SVG 1.1 element instance model).  If anything, it would move the `<use>` element closer to the web components shadow DOM implementation.  Independent clones means that the shadow DOM elements are no longer read-only from the perspective of author scripts, and no longer modified by the user agent in response to DOM mutations on the referenced graphics.

The only complication is defining at exactly what point the cloning occurs.  It would have to be at the later of: when the `<use>` element is fully parsed & added to the DOM, and when the referenced element is fully parsed & added to the DOM.

This is something for a new module (unless we have some really keen implementer responses), but I will be making sure the SVG 2 use element re-write supports the possibility.

progers:
Would it be possible to achieve the same goals by better supporting existing web components apis? A custom element extending from SVGElement is similar to a modern <use> but isn't possible today due to spec restrictions.

AmeliaBR:
@progers 

I'm still open to the idea of implementing custom elements in SVG, and I think others on the WG are, too.  However, this is a more limited proposal: no custom elements with unique functionality, just a shadow DOM of regular SVG elements doing what they usually do.

The benefit of implementing it as an extension to `<use>` is that you would still get the basic non-interactive cloning behavior in user agents or security situations that don't support scripting.  So your markup could define repeated objects that display even in a static image, but in interactive environments, each instance could react independently.

[...]

PR #206, opened 2016-07-13
Use element re-write & related changes

### LINKED ISSUE #90: Update href attribute descriptions to reference HTML51's obtain the resource algorithm
Two elements that use href, have prose defining how to fetch the resource.
This should probably be replaced by a link to [obtain](http://www.w3.org/TR/html51/document-metadata.html#obtain).

The two are:

The [script element](https://svgwg.org/svg2-draft/interact.html#ScriptElement).
> To get the referenced resource, do a potentially CORS-enabled fetch of the absolute URL that resulted from the earlier step, with the mode being the current state of the element's ‘crossorigin’ attribute, the origin being the origin of the ‘script’ element's Document, and the default origin behaviour set to taint. For origin, the rules defined for the html 'script' element apply the same way to the svg ‘script’ element. 
Related to Issue #88 

The [image element](https://svgwg.org/svg2-draft/embedded.html#ImageElement).
> To get the referenced resource, do a potentially CORS-enabled fetch of the absolute URL that resulted from the earlier step, with the mode being the current state of the element's ‘crossorigin’ attribute, the origin being the origin of the ‘image’ element's Document, and the default origin behaviour set to taint. For origin, the rules defined for the html 'img' element apply the same way to the svg ‘image’ element.

### LINKED ISSUE #99: We should define the behavior of ‘use’ in terms of Web Components
See https://svgwg.org/svg2-draft/struct.html#issue28
We should define the behavior of ‘use’ in terms of Web Components. See ACTION-3729 (Tab)

http://www.w3.org/Graphics/SVG/WG/track/actions/3729

### LINKED ISSUE #142: Clarify what is the size of a use element if width and height are not specified
If a ‘use’ element without 'width' and 'height' specified references a ‘symbol’ element, should the width and height of the ‘symbol’ element be used?

[The spec](https://svgwg.org/svg2-draft/struct.html#UseReUse) says: “If attributes ‘width’ and/or ‘height’ are not specified, the generated ‘svg’ element will use values of '100%' for these attributes.”

Does it mean:
1. if ‘width’ and ‘height’ are not specified _on the ‘use’ element_
2. if ‘width’ and ‘height’ are not specified _on the generated ‘svg’ element (after deep-cloning the ‘symbol’ element, which includes cloning its ‘width’ and ‘height’ attributes)_

Here’s an example: [https://jsfiddle.net/t2ps9huz/](https://jsfiddle.net/t2ps9huz/). The ‘use’ element doesn’t have width or height, but the ‘symbol’ element has. The browsers behave in two different ways.

symbol-pic

Option 1: Safari / Edge - uses 100% x 100%, ignores the symbol ‘width’ and ‘height’
Option 2: Firefox / Chrome - uses the symbol ‘width’ and ‘height’ (50x50)

It would be nice to know which one is the correct behavior so that I can file bugs for the browsers. I would prefer option 2 to be the correct behavior, because it would allow defining default width and height values for symbols.

The example as an attachment: 
[symbols.svg.txt](https://github.com/w3c/svgwg/files/270428/symbols.svg.txt)

### LINKED ISSUE #183: Define whether <use> shadow tree is interactive
SVG 1.1 allowed re-used "element instances" to be the target of events, with listeners set on the original elements responding to them, or with events bubbling up to the `<use>` element.  Since the SVG 1.1 element instance model was so inconsistently implemented, it's not really used in practice.  Furthermore, SVG 1.1 had no consideration of keyboard focusability & tabindex so did not address that aspect.  

<del>Also, no guidelines were given for whether interactive SMIL animations and CSS pseudoclasses applied to the source and re-used instance independently, or if they should be live mirroring.</del> Animations on the referenced element were required to be reflected in the re-used instance, which would presumably also apply to interactive effects, but it was not clear whether the instances could reflect interactive styles themselves.

Current browsers are currently a mess of inconsistencies when the re-used content inclues a link.  Results based [on this simple demo](http://output.jsbin.com/pufufagudo). The left circle is the original link, the right circle is the re-used link.  

- Firefox: both links are focusable and functional.  Focus styles are only applied to the correct element.

- Chrome: the re-used link is focusable and functional as a link, but focus styles are not applied, except sometimes as a mirror when the focus is actually on the original, so it's a usability mess.  (Aside: On my system, Chrome is not applying the fill colors to either circle. No idea why, but it seems to be caused by the `<a>` element not passing styles along, regardless of the `<use>` element.)

- Edge: The re-used link _appears_ to pointer users to be interactive (cursor changes, status bar shows link destination), but nothing happens on click and the link does not receive keyboard focus.

The current guidelines we have in the SVG-AAM assume that `<use>` shadow trees are inert, non-interactive, and therefore not exposed in the accessibility tree.  

By default with a Shadow DOM model, the content is interactive, with event-retargeting when events bubble out of the shadow and into the "light" `<use>` element.  If we want to over-ride this, we could define the shadow as an inert subtree.

### LINKED ISSUE #192: Use !important UA stylesheet rule to override display property for never-rendered elements
We currently have a lot of repeated prose in SVG 2 about how the `display` property doesn't apply to certain elements (the [never-rendered elements](https://svgwg.org/svg2-draft/render.html#TermNeverRenderedElement), in the terminology I added to the rendering chapter).

As I was looking up better ways to define the behavior of `<symbol>` from a Shadow DOM perspective (i.e., never-rendered directly, but rendered when re-used), I came upon the solution that it can be defined as an `!important` rule in the user agent stylesheet.

As per [CSS Cascading and Inheritance](https://www.w3.org/TR/css-cascade-4/#importance):

> User agent style sheets may also contain !important declarations. These override all author and user declarations.

In other words, instead of saying that `display` doesn't apply to these elements, we can just say that, no matter what CSS rules you apply elsewhere, a user agent `display: none !important` rule overrides them.

The net behavior is the same, but it's more logically consistent with CSS in general.  The only user-facing change is that if you `getComputedStyle()` for a never-rendered SVG element, you would now explicitly get `display: none` returned.

### LINKED ISSUE #193: Clarify effect of invalid values for width & height on <use> elements
The SVG 1.1 specs say, with regards to width + height on `<use>`:

> A negative value is an error (see Error processing). A value of zero disables rendering of this element.

Nothing very surprising about that. The same text is used for pretty much every other shape that takes `width` and `height` values.

But there are two problems:

- Any other value of `width` and `height` doesn't actually have any effect on the rendering of the re-used graphics if the `<use>` references anything that isn't `<svg>` or `<symbol>`.

- Current implementations are so inconsistent it's funny.

Demo for testing current browser implementations: http://jsbin.com/tixihodafe/1/edit?html,css,output
Or download [a zip of a stand-alone SVG version](https://github.com/w3c/svgwg/files/351182/use-width-height-error.zip)

For re-used basic shapes, Chrome, Edge, Safari, and Inkscape happily render the shapes regardless of width/height values.  However, Firefox follows the spec to the letter and does not render when width & height are 0 or negative.

For symbols, I would expect that invalid values of width & height mean no rendering.  In Firefox, once again that's true.  In Chrome, Edge and Inkscape, the 0 values effectively mean no rendering.  

But negative values are fun.  
Edge and Inkscape apply a negative scaling factor, so that the symbol gets reflected to the opposite side of its positioning point.  Chrome just ignores the negative value as if it wasn't specified (so that the symbol gets width/height  of 100%).  

Finally, Safari ignores _both_ the negative values and the zero values, using 100% in either case.

***So...***  my recommendation, and what I'm using in the text I'm drafting: 

Define valid values of width & height on `<use>` purely through their effect on the rendered content.  If the attributes don't have an effect (because the re-used content is a simple shape or a `<g>`), don't worry about error checking.  If the re-used content is a `<symbol>` or a `<svg>`, *and* it doesn't have its own width/height specified, then a value of 0 disables rendering, the same as it would if set on the symbol/svg itself.

(PS, I'm already writing it up so that width & height are valid style properties on `<symbol>`, as discussed in #142<del>, and so that the attributes on the `<use>` element only have an effect if the symbol/svg has a computed value of `auto` for the corresponding style property, regardless of how that property was set in the CSS cascade.</del>)

***But*** here's the final confusion: 

Because width & height on most elements have been upgraded to presentation attributes, and because CSS allows invalid negative values to be ignored *at parse time*, if you set a negative height/width directly on an `<svg>`, it should be ignored, and the auto/100% behavior would apply.

However, width & height on the `<use>` element are *not* currently defined as presentation attributes.  They are just regular SVG/XML attributes.  So, we can decide: are negative values ignored as if they were never set (so initial value applies) or are they an error that disables rendering?  I'm leaning towards ignored, so it's consistent with the presentation attribute behavior.  Also, if negative values aren't going to be a render-blocking error when re-using a `<g>`, they really shouldn't be a render-blocking error when re-using an `<svg>`.

If you're keeping score, that means that I'm suggesting that the Blink behavior would be correct, & everyone else would need to change something.

### LINKED ISSUE #203: Clarify meaning of context-fill & context-stroke keywords in paint definitions
The `context-fill` and `context-stroke` keywords are [listed as valid paint values](https://svgwg.org/svg2-draft/painting.html#SpecifyingPaint), but aren't <del>actually defined anywhere in SVG 2.</del> clearly defined at the point where they are used.

AmeliaBR:
## Overview

Re-write all text related to `<use>` element 
to be compatible with shadow DOM,
to define all the aspects of use elements that are currently undefined
(including those we'd added warnings about last year),
and to address various related issues.

To make it easier to review, I've pushed a built version of the changes to my GitHub Pages branch.  

**Major changes are in the section on the [Use element](https://ameliabr.github.io/svgwg/build/publish/struct.html#UseElement).**  

Other changes include:

- [Symbol element](https://ameliabr.github.io/svgwg/build/publish/struct.html#SymbolElement)
- [Symbol & Use-related interfaces](https://ameliabr.github.io/svgwg/build/publish/struct.html#InterfaceSVGSymbolElement)
- [rendering re-used graphics](https://ameliabr.github.io/svgwg/build/publish/render.html#ReusedGraphics) (and some other definitions tweaked in that chapter)
- [clarifying context keywords in "Specifying Paint"](https://ameliabr.github.io/svgwg/build/publish/painting.html#context-paint)
- [user agent style sheet now controls never-rendered elements](https://ameliabr.github.io/svgwg/build/publish/styling.html#UAStyleSheet); there are corresponding prose changes to the relevant element descriptions.
- New: [common guidelines for URL processing and external resource fetching](https://ameliabr.github.io/svgwg/build/publish/linking.html#processingURL)

See below & issues in the text for some remaining questions.  Please **do not** merge until the issues I've added have been removed from the source code.

## Breaking Changes from SVG 1.1

Many of these features were never implemented to the SVG 1.1 spec in Firefox; in Blink, some SVG 1.1 aspect were previously implemented but have since been changed (to be closer to the new shadow DOM model).

- Use element shadow trees now consist of actual SVGElement objects in a proper shadow DOM node tree, not SVGElementInstance pseudo-nodes.

- The root of the shadow tree is a ShadowRoot object.

- Events bubbling out of a use-element shadow tree must follow the shadow DOM re-targetting rules.

- Styles rules are resolved independently for clones versus the referenced graphics; in some cases (particularly re interactive pseudo-classes), this will result in different styles than the SVG 1.1 model.

- Some interactive animations are now triggered independently on the shadow content vs the referenced graphics.

-  Never-rendered elements will now have a computed style of `display: none`.  Rendering shouldn't change.

## Other new requirements & functionality

- The shadow tree must now have a ShadowRoot object, and implement extra functionality from shadow DOM specs.

- A new interface SVGUseElementShadowRoot is defined, inheriting from ShadowRoot, in case we want to add functionality later that is unique to use-element shadow trees.

- All elements in a subtree must now be cloned, not only graphics elements.

- The shadow tree has scoped stylesheets.

- Symbol elements can have x/y/width/height properties.

- SVGElement has been extended to replace the core functionality from SVGElementInstance.

- SVGSymbolElement now inherits from SVGGraphicsElement (so you can call `getBBox`, etc., on a symbol instance)

- <ins>A ShadowAnimation interface is defined, to represent WAAPI animations that need to be cloned to apply to an element instance, but must be kept in sync with their source.</ins>

## Clarified requirements (undefined or inconsistently implemented from SVG 1.1)

- External files are processed in secure static mode, so scripts don't run and references to additional resources are not followed _unless_ the elements that reference those resources are actually cloned into this document.  Same-document style blocks, however, _are_ processed, and those style rules are copied over and must apply in the shadow DOM scope.  Media queries in those copied stylesheets are interpretted in the main document's media.  Animations do not run in the external document, but they _are_ copied over to the shadow tree, where they operate in the main document's timeline.

- External files do not have to be .svg files, so long as the referenced element can be a child of an SVG container element.

- Percentage lengths in shadow DOM content are resolved based on the coordinate system into which they are rendered, not based on the source.  This is _consistent_ with SVG 1.1 spec text, but inconsistent with implementations.

- A use-element shadow tree must not be generated if the ‘use’ element is excluded because of conditional processing.

- Web Animations API animations must be duplicated, in addition to animation element.  CSS animations & transitions are covered under the general stylesheet duplication, and so are not duplicated directly.

- Animation elements that are triggered by an event on another element are also triggered by that event on an instance of the other element.

- Event listeners defined with `addEventListener` are copied to instances, the same as event attributes.

- <ins>secure processing mode is required for external resource files</ins>

- <ins>when a same-document URL reference to a resource can't initially be matched, the user agent should listen for DOM changes (e.g., new elements inserted) to find the match when available</ins>

## Clarifications or changes from previous drafts of SVG 2

- All aspects of use elements referencing external resources are now defined, so no warnings are required.  (Except there is still a warning re CORS not being supported, I haven't changed that.)

- refX, refY on `symbol` have an initial value of (none), different from a value of 0, to preserve backwards compatibility when the symbol has a viewBox.

- audio and video elements have additional requirements to maintain synchronization.

- A use element is _not_ automatically a stacking context.  Patterns, markers, etc. _are_ automatically isolation groups for blending.  Previous draft of the rendering chapter tried to apply the same rules for both cases.

- Role mappings for `symbol` and `use` have been changed to reflect that use-element shadow content must be fully accessible.

- The `instanceRoot` (and `animatedInstanceRoot`) IDL properties have been restored on `SVGUseElement`.  The animated version is just an alias, there's no requirement to maintain multiple shadow trees.

## Issues Addressed

When merged, this pull request:

- Closes #99, "We should define the behavior of ‘use’ in terms of Web Components" by defining use elements in terms of shadow DOM, which is related to but not the same as web components.

- Closes #142, "Clarify what is the size of a use element if width and height are not specified", by allowing symbol elements to have geometric properties (x/y/width/height) and applying the same rules as for re-used `<svg>`.

- Closes #193, "Clarify effect of invalid values for width & height on <use> elements", by making negative values parse errors and by defining the impact of 0 in terms of the shadow element.

- Closes #183, "Define whether <use> shadow tree is interactive" by making it fully interactive and accessible, but still read-only.

- Closes #192, "Use !important UA stylesheet rule to override display property for never-rendered elements", by doing so.

- Closes #203, "Clarify meaning of context-fill & context-stroke keywords in paint definitions", by making the cross-links and wording tweaks I discussed there.

- New: Closes #90 "Update href attribute descriptions to reference HTML51's obtain the resource algorithm" by making all resource fetching (via href or via style properties) reference HTML's potentially-CORS-enabled fetch request algorithm.

## Issues Remaining

<del>Most of these are noted in the text</del><ins>All resolved, with only a couple permanent issues re at-risk features</ins>:

- I would really, really prefer to have [x/y/width/height on `<use>` elements](https://ameliabr.github.io/svgwg/build/publish/struct.html#UseAttributes) be proper geometric properties, instead of simple attributes.  I just haven't quite figured out how that would conceptually work.  As is, it's going to be a surprise & a pain for authors who are going to want to control these properties with CSS.

  <ins>Resolved: Made the switch, as discussed. One way or the other, values on the use element override CSS properties on the shadow child; might as well make the use values convenient.</ins>

- Re [`defs` element](https://ameliabr.github.io/svgwg/build/publish/struct.html#Overview):  Is there any justification for continuing to recommend "that, wherever possible, referenced elements be defined inside of a ‘defs’ element"?

  <ins>Resolved: Removed/replaced with a gentler suggestion for good code organization</ins>

- Re [resolving cross-references](https://ameliabr.github.io/svgwg/build/publish/struct.html#UseHref):  Should this section, or parts of it, be moved to the Linking chapter, and made to apply for all external resources (e.g., external paint servers, filters, etc.)? Also:  Do we need to reference specifics re fetching an external document and security modes? Related to #90.

  <ins>Resolved: Section is moved & greatly re-written to make it apply to all external resources, including specifics re fetching & security</ins>

- Should there be some sort of load event that is dispatched after the use tree is successfully generated? If so, which event in particular?

  <ins>Resolved: Make this a future wish list item</ins>

- Does it make sense that conditional processing stop formation of a use-element shadow tree? Consistent with audio/video not playing with conditional processing (see #44), and with the idea that it stops all processing (including fetching external resources), not only rendering.  (`display: none` does not prevent shadow tree formation).

  <ins>Resolved: Conditional processing only affects display, not processing.</ins>

-  Is `NoModificationAllowedError` the best choice for the error type whenever something is blocked by the read-only nature of the shadow tree? (Other options include ReadOnlyError and NotAllowedError.) Needs to encompass both DOM modification attempts as well as attribute or property changes.

  <ins>Resolved: Keep it as written.</ins>

- Should width & height on a use element affect foreignObject and embedded content? The section on establishing a new viewport also includes foreignObject, iframe, and image, although then it qualifies that for image and iframe the new viewport actually is created by the root element of the embedded document. Mostly, that whole section needs re-writing. But for this section: should auto width/height on a re-used foreignObject be influenced by width/height on the use element? SVG 1.1 didn't allow foreignObject to be re-used, so it would not be a breaking change either way.

  <ins>Resolved: No, foreignObject doesn't behave like svg/symbol element; referenced section clarified to distinguish between different types of viewports & coordinate system resets.</ins>

- Should the details on how animations work in use shadow trees be moved to the SVG Animations specification and/or WAAPI? Most of the details are dependent on those specs. However, implementers that currently support SVG animation elements and Web Animations are going to need guidance on how to update those implementations to match the new shadow DOM model. Also, it would be nice to have a TR version of SVG Animations to link to!

  <ins>Resolved: Keep section here. </ins>

- Is it going to be an implementation hassle to have detached Element objects (cloned animation elements that affect the shadow tree but aren't cloned as part of it) that must behave as if they are part of the node tree? Another option would be to actually make them additional child nodes of the ShadowRoot node. I don't think that would have any problematic side effects, although I'd need to change the wording a few places where I specify that the use-element shadow root only ever has one child.

  <ins>Resolved: Keep it as written, unless someone complains when implementing.</ins>

- Is distinguishing between begin="target.event" and begin="event" worth the extra implementation complexity? It creates useful functionality and is consistent with the CSS model, but it is technically a breaking change from SVG 1.1.

  <ins>Resolved: Keep it as written.</ins>

- Re [SVGElementInstance mixin-interface](https://ameliabr.github.io/svgwg/build/publish/struct.html#InterfaceSVGElementInstance): Is it appropriate to define this as a [NoInterfaceObject] interface? As defined in this spec, the interface is only used to supplement SVGElement. But SVG 1.1 scripts could be explicitly testing for the presence of Window.SVGElementInstance to determine if the browser supports an inspectable use-element shadow DOM.

  <ins>Resolved: Keep as written.</ins>

- Re [SVGElementInstance::correspondingElement](https://ameliabr.github.io/svgwg/build/publish/struct.html#__svg__SVGElementInstance__correspondingElement): Should this still point to a valid element object even if it is in a different document? That implies that the entire DOM of the external file must be maintained in memory (since the element linked from this property would be linked to all other nodes in its DOM tree). I've already said that that external DOM is read-only, so there's no real function to having it available.

  <ins>Resolved: Keep as written, but add explanation & at-risk warning.</ins>

## Still To Do

- Re-define paint server cross references in terms of shadow DOM trees. (see #121)

- <del>Add new examples of `use` elements in action.</del>

- Update the changes appendix.


master/struct.html
```diff
@@ -729,13 +777,13 @@ <h3 id="SymbolAttributes">Attributes</h3>
       <tr>
         <td><dfn data-dfn-type="element-attr" data-dfn-for="symbol" id="SymbolElementRefXAttribute">refX</dfn></td>
         <td><a>&lt;length&gt;</a> | left | center | right</td>
-        <td>0</td>
+        <td>(none)</td>
         <td>yes</td>
       </tr>
       <tr>
         <td><dfn data-dfn-type="element-attr" data-dfn-for="symbol" id="SymbolElementRefYAttribute">refY</dfn></td>
         <td><a>&lt;length&gt;</a> | top | center | bottom</td>
```
nikosandronikos: Do we need to define what the keywords evaluate to here?

master/struct.html
```diff
@@ -905,49 +1028,303 @@ <h3 id="UseAttributes">Attributes</h3>
...
+  A shadow root should be implemented in conformance with the <a href="https://www.w3.org/TR/shadow-dom/">Shadow DOM</a> specification <a href="refs.html#ref-SHADOWDOM">[SHADOWDOM]</a>,
+  or its future replacement.
+  </dd>
+  
+  <dt id="TermShadowHost">shadow host</dt>
+  <dt>host</dt>
+  <dd>An element that has an associated shadow root.</dd>
+  
+  <dt id="TermShadowTree">shadow tree</dt>
+  <dd>A node tree whose root is a shadow root.</dd>
```
nikosandronikos: Can we just link to the DOM definition of a shadow tree?
https://dom.spec.whatwg.org/#shadow-trees

nikosandronikos:
> I would really, really prefer to have x/y/width/height on <use> elements be proper geometric properties, instead of simple attributes. I just haven't quite figured out how that would conceptually work. As is, it's going to be a surprise & a pain for authors who are going to want to control these properties with CSS.

Is the only issue the difference in initial values for width and height?

nikosandronikos:
> Re defs element: Is there any justification for continuing to recommend "that, wherever possible, referenced elements be defined inside of a ‘defs’ element"?

If you can't think of any then I highly doubt it, since the justification is accessibility and understandability. I'm happy for you to remove it.

nikosandronikos:
> Re resolving cross-references: Should this section, or parts of it, be moved to the Linking chapter, and made to apply for all external resources (e.g., external paint servers, filters, etc.)? Also: Do we need to reference specifics re fetching an external document and security modes? Related to #90.

This seems related to #90 
It would be nice to have in one place. I'd like to get some review from the HTML and DOM editors so maybe leave it for now and we'll see what they recommend.

nikosandronikos:
> Should there be some sort of load event that is dispatched after the use tree is successfully generated? If so, which event in particular?

Sounds useful, but has it been discussed before? I say file a new issue tagged 'future wish list' for now.

nikosandronikos:
> Does it make sense that conditional processing stop formation of a use-element shadow tree? Consistent with audio/video not playing with conditional processing (see #44), and with the idea that it stops all processing (including fetching external resources), not only rendering. (display: none does not prevent shadow tree formation).

Making it function just like one would use a MQ seems like the best approach to me. Treat it as display:none and create the shadow tree.

nikosandronikos:
> Is NoModificationAllowedError the best choice for the error type whenever something is blocked by the read-only nature of the shadow tree? (Other options include ReadOnlyError and NotAllowedError.) Needs to encompass both DOM modification attempts as well as attribute or property changes.

NoModificationAllowedError is consistent with read only properties on other SVG DOM objects so I think it's the best choice.

nikosandronikos:
> Should width & height on a use element affect foreignObject and embedded content? The section on establishing a new viewport also includes foreignObject, iframe, and image, although then it qualifies that for image and iframe the new viewport actually is created by the root element of the embedded document. Mostly, that whole section needs re-writing. But for this section: should auto width/height on a re-used foreignObject be influenced by width/height on the use element? SVG 1.1 didn't allow foreignObject to be re-used, so it would not be a breaking change either way.

Is this just about what's useful to authors? It sounds like we should use width and height on the use.

AmeliaBR:
Thanks for all the comments @nikosandronikos!

Re  x/y/width/height as simple attribute versus presentation attributes for the corresponding geometric properties:

The issue is that it can't be described with either the CSS inheritance model or with a container model.  The used value for the "child" (`<svg>` or `<symbol>`) element is the value set on the "parent" (`<use>`), _iff_ the parent has a non-auto value, otherwise use the child's own computed value (or if it's also auto, use 100%).  There's no way to define that with user agent stylesheet rules, or with rules for the conversion of `auto` into lengths.

However, at this point I'm inclined to make them presentation attributes anyway.

One way or the other (simple attributes or presentation), we end up with special over-ride rules where the used value for the `<svg>`/`<symbol>` isn't based on their computed value.  So we might as well add in the convenience of being able to set the values on the `<use>` with CSS.

AmeliaBR:
Re `foreignObject` and width/height:

> Is this just about what's useful to authors? It sounds like we should use width and height on the use.

It's more about what is consistent / clear.  An `<image>` doesn't adjust its size based on the `<use>` width & height. Neither does a `<rect>`.

Since a `<foreignObject>` doesn't take a `viewBox` attribute, I'm leaning towards not including it in the size-to-fit effect.  If & when we re-define `viewBox` as a presentation attribute that can apply on any element, then we can change the `<use>` behavior to kick in whenever a new viewport is created in the child content.

So the change is really a matter of cleaning up the section on viewport-creating elements to distinguish between the different cases. A `<foreignObject>` doesn't create a new SVG coordinate system, anyway: it creates a new CSS box model container.

AmeliaBR:
Re conditional processing:

> Making it function just like one would use a MQ seems like the best approach to me. Treat it as display:none and create the shadow tree.

This is the easiest approach to deal with conditional processing, but it's also the least useful.  It would mean that the answer to #44 is that no, conditional processing does _not_ stop audio from playing (since `display: none` does not).  Quite frankly, the neutering of conditional processing started when the WG agreed that scripts and stylesheets wouldn't be affected by it, so this is just a final nail.  But it would be unfortunate.  What's the point of doing feature-support testing if you still end up loading all your fallback external resources in either case?

I would much prefer to have conditional processing make the sub-tree completely inert, like an HTML `<template>`.  That would mean re-visiting the previous resolution about `<script>`.

master/struct.html
AmeliaBR: The DOM specs are a bit in flux, so I was hesitant to just link.  But maybe should link in addition to defining them here.

master/struct.html
AmeliaBR: Yes, probably just with a link to the equivalent definitions for `<marker>`.

Tavmjong:
The change in styling behavior is waving a red flag in front of my eyes. Can you expand on this? We (Inkscape) make heavy use of the <use> element. An example would be nice.

AmeliaBR:
> The change in styling behavior is waving a red flag in front of my eyes. Can you expand on this? We (Inkscape) make heavy use of the element. An example would be nice.

The intention is that most cases won't change, except for things like `:hover` effects, which are currently a mess in browsers.  But an example of a case that _would_ change is as follows:

```
<svg>
  <style>
    svg {stroke-width: 10}
    .special {fill: blue}
    .special circle {stroke: green}
    use {fill: orange; stroke: purple}
  </style>
  <g class="special">
     <circle id="c" cy="50" cx="50" r="40" />
  </g>
  <use xlink:href="#c" x="100" />
</svg>
```

([JSBin demo](http://jsbin.com/pudivoweke/edit?html,output))

In a user agent that strictly conformed to SVG 1.1, the re-used circle would be orange with a green stroke.  The original `<circle id="c">` element directly matches the `.special circle` rule, so it gets a green stroke, but it's fill value is `inherit`.  Those two values are then copied to the circle element instance created by the `<use>`, and the inherited fill value is then computed as orange.

Under the new model, the copied circle would be orange with purple stroke.  The shadow DOM `<circle>` doesn't have an ancestor with class "special", so it doesn't match that rule. It inherits both fill and stroke.

For what it's worth, Firefox's broken style matching model already breaks this (it draws the re-used circle with a purple stroke).  So I'm not breaking anything that works cross-browser on the web.  I'm also pretty sure that Inkscape and other WYSIWYG software aren't creating descendent-selector CSS rules like `.special circle` by default, so it's not likely to affect content generated by the software internally.

AmeliaBR:
PS, I could add this example to the note warning about the change, if that would be helpful.

@Tavmjong

nikosandronikos:
> >     Should there be some sort of load event that is dispatched after the use tree is successfully generated? If so, which event in particular?
> Sounds useful, but has it been discussed before? I say file a new issue tagged 'future wish list' for now.

I've filed this as #214

adanilo:
I really should have paid more attention to this earlier, apologies. Anyway, this is a major breaking change from 1.1, namely specifying an actual clone. I get that aligning with Shadow DOM is a nice idea, but <use> was always supposed to be a 'effective' clone, not an actual one. i.e. it's supposed to be a DAG primitive, not a templating mechanism. As far as I recall, Ikivo, Presto, Abbra and I believe Edge use a DAG approach for rendering. A large number of bugs in webkit/Blink/Gecko are unfortunate side-effects from implementing an actual cloning mechanism, and then trying to keep the clones all in sync when the referenced tree changes. It'd be nicer to have a new element if you want to create true clones IMO, or just not break it:-)

birtles:
I haven't followed the discussion on this topic so I didn't know we were actually exposing the shadow tree. Assuming that is the approach we want going forward, then I think the animation section here seems reasonable.

However, with regards to how we spec the Web Animations part, I think that rather than defining everything in prose, it would be preferable to introduce a new sub-type of `Animation` that exhibits the mirroring behavior. e.g.

```webidl
[Constructor (Animation source)]
interface ShadowAnimation : Animation { };
```

Then we could define that `ShadowAnimation` objects throw `NoModificationAllowedError` (or whatever the appropriate error is in this case) for each of the mutator methods (e.g. `pause()` etc.).

Furthermore, you might expose the `source` `Animation` as a readonly attribute and further spec that this object reflects all changes to `source`.

So, I guess what I'm suggesting is to define `ShadowAnimation` as a generic thing that could be potentially used in other situations, and then just use that concept here.

The idea of adding sub-types of `Animation` is not new. The [`CSSAnimation`](https://drafts.csswg.org/css-animations-2/#cssanimation) and [`CSSTransition`](https://drafts.csswg.org/css-transitions-2/#csstransition) interfaces are already defined elsewhere and we're [currently looking at defining a `FillAnimation` sub-interface](https://github.com/w3c/web-animations/issues/118) that would also throws on most mutation operations--it's slightly different, however, in that it still allows cancelling.

The interaction between `FillAnimation` objects and `ShadowAnimation` objects would need to be defined, but first we need to investigate if `FillAnimation` is going to happen or not.

AmeliaBR:
@adanilo 

I appreciate the input, but I think that it's a little late to reverse course.  Certainly, you'd need to convince the Blink team, who have already moved ahead unilaterally with many of the Shadow DOM changes.

Quite frankly, with the movement of geometry to CSS, including the potential to inherit a variable that changes geometry, I don't see much performance benefit in _not_ having distinct clones.  All rendering & layout information will need to be calculated independently for the instances.  Event handling and interaction was also always supposed to be independent.

I recognize that the remaining mirroring behavior is still a pain, but it should help that dynamic style changes will no longer need to be propagated.  Similarly, animations will be mirrored by mirroring the original animation instructions, rather than by propagating animated values.  So this should make it easier to run animations or any type on independent threads.  Only actual DOM changes, new event listeners, and new animation objects will force the two trees to sync.

I've suggested elsewhere ( #184 ) that I'd like to add a toggle attribute to `<use>` elements to create unlinked, independent clones, but that will have to wait for a future spec.

AmeliaBR:
@birtles 

The `<use>` tree was always open in the SVG spec.  It has just never got a lot of attention in practice because Firefox didn't implement it that way!

I like your suggestion of sub-classing `Animation` to define the mirrored copies.

Would you be open to using `correspondingAnimation` instead of `source`, to mirror the SVG terminology? Or is that uncomfortably verbose for a feature that you hope to have wider use?

birtles:
> Would you be open to using `correspondingAnimation` instead of source, to mirror the SVG terminology? Or is that uncomfortably verbose for a feature that you hope to have wider use?

@AmeliaBR It's pretty verbose and remembering the double 'r' in corresponding is going to trip up non-native English speakers everywhere. On the other hand, `source` might be a bit too ambiguous, so `sourceAnimation` ?

adanilo:
Hi @AmeliaBR yes, I know it's too late, that's why I apologized for not paying attention. Performance we can speculate about, but memory is another thing altogether. With the DAG model, if I design a nice intricate graphic (e.g. a snowflake) with 1,000 nodes and reference it 1,000 times in my snow village, I get 2,000 DOM nodes of memory consumed vs. 1,001,000 which is fairly considerable. It's just a bit strange to formalize a 'mistake' in some implementations... In any case, event handling from an implementation that implements the DAG correctly doesn't seem to have caused existing implementers any difficulty in the past. I agree this will have little effect on rendering. Anyway, we'll see what implementers do over time. I do like the concept of bringing Shadow DOM into SVG, just necessarily via <use> which already worked IMHO.

PR #251, opened 2016-08-27
Add text-transform property to the list of required CSS properties

BigBadaboom:
Since all browsers already seem to support this property (I tested Chrome, FF and IE11), I figured we might as well add this to the list of "Required properties" in  Styling section 6.7.

https://svgwg.org/svg2-draft/styling.html#RequiredProperties

### DIFF
--- a/master/styling.html
+++ b/master/styling.html
@@ -461,7 +461,6 @@ the presentation attribute name is the same as the property name, in lower-case 
       <a>'text-decoration'</a>,
       <a>'text-overflow'</a>,
       <a>'text-rendering'</a>,
-      <a>'text-transform'</a>,
       <a>'unicode-bidi'</a>,
       <a>'vector-effect'</a>,
       <a>'visibility'</a>,
--- a/master/styling.html
+++ b/master/styling.html
@@ -461,6 +461,7 @@ the presentation attribute name is the same as the property name, in lower-case 
       <a>'text-decoration'</a>,
       <a>'text-overflow'</a>,
       <a>'text-rendering'</a>,
+      <a>'text-transform'</a>,
       <a>'unicode-bidi'</a>,
       <a>'vector-effect'</a>,
       <a>'visibility'</a>,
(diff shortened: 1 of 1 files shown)


BigBadaboom:
Test fiddle: https://jsfiddle.net/t06dagvj/

AmeliaBR:
@BigBadaboom My concern would not be the browsers, since as you say they all support it already, but other software.  Text-transform isn't easy to do right (you need different rules for different languages).

@Tavmjong, do you have any opinions on this? Does Inkscape support text-transform? Does anyone know whether Adobe does?

AmeliaBR:
PS, Just because a property isn't _required_ doesn't mean it isn't _compatible_ with SVG.  I can't recall whether we already say that clearly in the Styling chapter or not.

BigBadaboom:
> Just because a property isn't required doesn't mean it isn't compatible with SVG. I can't recall whether we already say that clearly in the Styling chapter or not.

I don't remember seeing any copy to the effect.

>  Text-transform isn't easy to do right. 

That's a fair point.  I'm sensitive to that argument myself because of my library.  However most OSes have support (or at least libraries) for doing correct capitalisation transforms now.  And going by the table on MDN, none of the browsers do it all completely correct anyway.

https://developer.mozilla.org/en/docs/Web/CSS/text-transform

BigBadaboom:
BTW, this PR is mostly a suggestion patch. If people think it is not necessary, I am not going to die fighting for it. :)

AmeliaBR:
> BTW, this PR is mostly a suggestion patch.

Understood & appreciated. Makes it easy to add it if a decision is made, but I didn't want to add it without checking with the implementers who would be suddenly made non-conforming.

I checked, and Inkscape has a "Change Case" extension, but it applies a one-time manipulation to the actual text data; the style property isn't supported.  I also don't know how well it handles languages, but I do know it depends on some external library I don't have installed.

Tavmjong:
Inkscape does not handle language specific rules when transforming case in the "Change Case" extension. There is a glib function that does handle case transforms respecting locale so it might not be too hard to support 'text-transform' (how good it is, I don't know). However, it is not quite as simple as adding 'text-transform' to the list of required properties as transforming text can change the number of 'typographic characters' as in the case of transforming the German lower case 'ß' to the uppercase 'SS'. The assignment of 'dx', etc. attribute values to 'typographic characters' should probably be handled as with ligatures.

The spec already states that: "Other CSS properties that affect text layout and rendering may also be supported by an SVG user agent; their effect should be taken into account as part of the CSS text layout step of the overall SVG text layout process." So I'm not too inclined to add 'text-transform' to the list of required properties at this point given the possible complications and the inconsistent support by browsers..

plehegar:
Marked as non-substantive for IPR from ash-nazg.

boggydigital:
Please see the GitHub-bot comments - we've resolved that we won't make this property required.

PR #294, opened 2016-11-15
Copy Editing: Eradicate the usage of informal `you'; clean up surrounding text

mfwitten:
(Ultimately, this commit should not change the meaning of the text.)

Because this is a standard, it is probably best not to employ the rather
informal second person, `you'. This commit removes that usage without
descending into overtly bureaucratic language.

Furthermore, as a matter of practicality or even necessity, some of
the nearby text has also been improved (according to this author);
this includes rewording, re-styling, the removal of typos, and the
fixing of grammar.

### DIFF
--- a/master/embedded.html
+++ b/master/embedded.html
@@ -457,17 +457,19 @@ or for a <a>'canvas'</a> element if scripting is disabled.
 
 <edit:example href='images/embedded/sourceinsvg.svg' name='sourceinsvg' link='no' image='no'/>
 
-<p class="note">
-  Currently, the HTML parser does not recognize these tagnames within an SVG subtree 
-  to be HTML-namespaced elements, although this may change in the future. 
-  Therefore, in order to include these elements within your SVG, 
-  you must either use an XML serialization that recognizes namespace declarations 
-  (such as stand-alone SVG or XHTML),
-  or you must create namespaced elements via the <code>createElementNS</code> DOM API method. 
-  Alternatively, you can use the foreignObject element 
-  to wrap the HTML-namespaced elements; 
-  the HTML parser correctly parses HTML elements within a foreignObject.
-</p>
+<div class="note">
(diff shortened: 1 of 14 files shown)


master/shapes.html
```diff
@@ -256,7 +256,7 @@ <h2 id="EllipseElement">The <span class="element-name">'ellipse'</span> element<
 <p>
   The <a>'cx'</a> and <a>'cy'</a> coordinates define the center of the ellipse.
 </p>
-
+</bx
```
SebastianZ: This is obviously a syntax error.

Sebastian

mfwitten: *Woops!* How embarrassing. I must have introduced that error at the very last moment; clearly, I need to figure out how to introduce an XML validator into my workflow.

As this is an obvious and trivial mistake, I've force-pushed an amended commit.

Thanks!

master/embedded.html
```diff
@@ -457,17 +457,19 @@ <h2 id="HTMLElements">HTML elements in SVG subtrees</h2>
...
-  Therefore, in order to include these elements within your SVG, 
-  you must either use an XML serialization that recognizes namespace declarations 
-  (such as stand-alone SVG or XHTML),
-  or you must create namespaced elements via the <code>createElementNS</code> DOM API method. 
-  Alternatively, you can use the foreignObject element 
-  to wrap the HTML-namespaced elements; 
-  the HTML parser correctly parses HTML elements within a foreignObject.
-</p>
+<div class="note">
+  <p class="note">
```
mfwitten: I should probably remove this unnecessary `class="note"`, as it is already specified by the parent element.

nikosandronikos:
Marked as non-substantive for IPR from ash-nazg.

nikosandronikos:
This improves the text in many places. Thanks!

PR #295, opened 2016-11-16
Make SVGUnitTypes a regular interface with only constants

### LINKED ISSUE #291: Values in SVGZoomAndPan and SVGUnitTypes should be accessible to javascript
Please:
The [SVG 2 specification](https://www.w3.org/TR/2016/CR-SVG2-20160915/types.html#InterfaceSVGZoomAndPan)

Makes it impossible to write var x = ZoomAndPan.SVG_ZOOMANDPAN_UNKNOWN as ZoomAndPan is [NoInterfaceObject].

Similarly with SVGUnitTypes.

This is not what either Chrome or Firefox do (both support accessing the ZoomAndPan interface values directly). In fact Firefox has tests that contain SVGUnitTypes so it's certainly likely that there's also usage of this in the wild.

I propose changing the ZoomAndPan interface to

```
interface SVGZoomAndPan {
};

SVGZoomAndPan implements SVGZoomAndPanValues;
```

Then having this as SVGZoomAndPanValues

```
[NoInterfaceObject]
interface SVGZoomAndPanValues {

  // Zoom and Pan Types
  const unsigned short SVG_ZOOMANDPAN_UNKNOWN = 0;
  const unsigned short SVG_ZOOMANDPAN_DISABLE = 1;
  const unsigned short SVG_ZOOMANDPAN_MAGNIFY = 2;

  [SetterThrows]
  attribute unsigned short zoomAndPan;
};
```

And all existing users of ZoomAndPan i.e. SVGSVGElement and SVGViewElement would then implement SVGZoomAndPanValues

This would mean that an SVGSVGElement would not be an SVGZoomAndPan but you would be able to access its contents.

Similarly for SVGUnitTypes.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1241899 and https://bugzilla.mozilla.org/show_bug.cgi?id=1241898 for the Firefox implementation which is landing now. This will align Firefox with Chrome's implementation.

foolip:
Fixes #291.

### DIFF
--- a/master/pservers.html
+++ b/master/pservers.html
@@ -3465,7 +3465,6 @@ for gradient paint server element interfaces.</p>
 };
 
 <a>SVGGradientElement</a> implements <a>SVGURIReference</a>;
-<a>SVGGradientElement</a> implements <a>SVGUnitTypes</a>;</pre>
 
 <p>The numeric spread method type constants defined on <a>SVGGradientElement</a>
 are used to represent the keyword values that the <span class='attr-name'>'spreadMethod'</span>
@@ -3638,7 +3637,6 @@ in the DOM.</p>
 
 <a>SVGPatternElement</a> implements <a>SVGFitToViewBox</a>;
 <a>SVGPatternElement</a> implements <a>SVGURIReference</a>;
-<a>SVGPatternElement</a> implements <a>SVGUnitTypes</a>;</pre>
 
 <p>The <b id="__svg__SVGPatternElement__patternUnits">patternUnits</b>
 and <b id="__svg__SVGPatternElement__patternContentUnits">patternContentUnits</b>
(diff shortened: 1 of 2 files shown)


foolip:
Must I be a WG member to write a PR? I'd rather not join for a one-off like this.

AmeliaBR:
@foolip You don't have to be a WG member, but you do have to click OK to an IP agreement, acknowledging that your code & words are being freely contributed, etc.

PS, Thanks for writing this up.

foolip:
@AmeliaBR, where can I click "OK"? I tried following the "Details" link for the failed check but it looks like an admit view for the repo owner or something.

nikosandronikos:
@foolip: Are you at Opera?
I think the issue is that Opera doesn't have any affiliation with the SVG WG anymore (maybe you should?).
My understanding is that the ash-nazg tool would allow contributions from you without you being a WG member, as long as you have a single Member-affiliation and that member is in the group.

I'll follow this up with @shepazu and get back to you. It may be that someone from Opera needs to join the WG, or you need to fill out one of [these forms](https://www.w3.org/2004/01/pp-impl/83482/nmlc).

foolip:
@nikosandronikos no I'm at Google now.

plehegar:
(ip check solved. we just got an improvement earlier today to avoid failing on this type of cases)

PR #304, opened 2017-02-10
Remove explicit currentColor from <paint>

Manishearth:
According to CSS 3 Color, `currentColor` is a part of `<color>`, so we don't need to explicitly specify it

r? @nikosandronikos


dirkschulze:
Trivial change that doesn't need accepting IP.

Issue #321, opened 2017-05-27
Serialization format for path strings


ewilligers:
This issue came up in an Intent to Ship (offset-path animation) discussion on blink-dev.
https://groups.google.com/a/chromium.org/d/msg/blink-dev/_DPl-JG6bV8/y2_GUwFVDwAJ

'd' path strings can contain a mix of absolute and relative coordinates, and a mix of comma and 
space separators. https://www.w3.org/TR/SVG2/paths.html#TheDProperty

It is also possible to animate between paths that differ in the use of separators and absolute and
relative coordinates, for example between 'M 200 400 H 500 v -100' and 'm100,200h400V300'.

Now that 'd' is exposed as a presentation attribute, page authors can use getComputedStyle to
inspect the path string.

For consistency between implementations, it would be helpful if the standard specified absolute
[or relative] coordinates in the serialization format, and space separators.

The above example path strings would become 'M 200 400 H 500 V 300' and 'M 100 200 H 500 V300'
when serialized.

Blink's currently shipping implementation uses space separators, and uses the 'to' format for
absolute/relative coordinates, except at animation progress 0, where the 'from' format is used:-
property <d> from [path('M 200 400 H 500 v -100')] to [path('m100,200h400V300')]
 at (0) is   [path('M 200 400 H 500 v -100')]
at (0.5) is [path('m 150 300 h 350 V 300')]
at (1) is   [path('m 100 200 h 400 V 300')]

If we specify absolute coordinates for the serialization format, these would become:
property <d> from [path('M 200 400 H 500 v -100')] to [path('m100,200h400V300')] 
at (0) is   [path('M 200 400 H 500 V 300')]
at (0.5) is [path('M 150 300 H 500 V 300')]
at (1) is   [path('M 100 200 H 500 V 300')]

If we specify relative coordinates for the serialization format, these would become:
property <d> from [path('M 200 400 H 500 v -100')] to [path('m100,200h400V300')] 
at (0) is   [path('m 200 400 h 300 v -100')]
at (0.5) is [path('m 150 300 h 350 v 0')]
at (1) is   [path('m 100 200 h 400 v 100')]

No backward compatibility issues are anticipated in moving to consistent absolute coordinates [or consistent relative coordinates].

If issue #320 is accepted, there is also the choice of quote character around the path string. Blink uses single quotes.

ewilligers:
There was an additional suggestion from Brian Birtles
[...]

Issue #119, opened 2016-04-21
Make d a presentation attribute on path


nikosandronikos:
Resolved that for now d will become a presentation attribute (no name change) with path data string as a value and for now it only affects the path element.

https://www.w3.org/2016/04/21-svg-minutes.html#resolution03

Need to update https://svgwg.org/svg2-draft/paths.html#DProperty

Tavmjong:
Removed need for path() in 'd' value. Commit 0753aed5f7932f03123d3d94fb2533d402ea4455
Should complete making 'd' a property.

SelenIT:
It seems that the `d` property is also missing from [the property index](https://svgwg.org/svg2-draft/propidx.html), should it be added there?

Issue #49, opened 2016-02-18
`d` geometric property needs a clear & extensible name


AmeliaBR:
The edits to make path data a presentation attribute currently use the same `d` property name for both the shape of a `<path>` element and the baseline shape of a `<textPath>`.  It is not currently used for other elements (such as `<hatchpath>` and `<stop>` elements in a `<mesh>`) that use path data with the `path` attribute.

This creates some concerns:

- `d` is not a particularly expressive property name outside the context of a particular element.

- the same `d` property is being used to apply two very different features, with only the element tag name determining how (or if) the property has an effect.

- the `points` attribute of `<polygon>`/`<polyline>` elements, which is a natural equivalent to `d` on a `<path>`, is not currently reflected in CSS geometric properties.

_____________________________________

Main discussion on the www-svg mailing list starts here:

- https://lists.w3.org/Archives/Public/www-svg/2016Feb/0013.html

Also relevant is this discussion on the FX mailing list, regarding the `path()` function syntax, which unfortunately got broken into two threads, only the shorter of which was copied to www-svg:

- https://lists.w3.org/Archives/Public/public-fx/2016JanMar/0027.html
- https://lists.w3.org/Archives/Public/public-fx/2016JanMar/0023.html

AmeliaBR:
Some rough sketches of a proposal.

See also the discussion from the [3 March 2016 telcon](https://www.w3.org/2016/03/03-svg-minutes.html#item02).

I think there is agreement that we need distinct property names for distinct features:

- text path
- geometry of paths & probably other shapes
- _maybe_ geometry path fragments used within the new paint servers (or maybe they could be covered under same property as geometry of paths)

I've written up notes on text path as Issue #59.  This issue will focus on shape geometry.

## Defining shape geometry in CSS

- There is a desire from the CSS side to use CSS shape functions instead of raw path data strings as the value of the property.

- This already means that there won't be a simple presentation attribute which matches the property syntax.  So might as well create a better property, which then pulls in the value of the attribute through the user stylesheet.

- Given this, it should be straightforward to use the same property for path/polygon/polyline.  Just use different shape functions and different `attr()` calls in the default stylesheet.  We would need to define a new `polyline()` function, but that will be useful for CSS motion path as well.

- I had suggested a `geometry: <shape-function>` as the simple-to-understand syntax.

[...]

PR #374, opened 2017-12-18
'd' presentation attribute: use path() function

### LINKED ISSUE #320: 'd' presentation attribute inconsistent with shipped implementation and offset-path
Blink shipped the d presentation attribute with value: path(<string>) | none
based on the Sydney face to face discussions early in 2016, and the spec text
from around the same time.

This is consistent with the syntax for offset-path
https://drafts.fxtf.org/motion-1/#offset-path-property

I didn't realize the syntax had changed in a teleconference
https://github.com/w3c/svgwg/issues/119
until https://crbug.com/652822 was raised after Blink shipped.

It isn't possible to change the syntax returned by getComputedStyle in a backwards
compatible way.

Please consider restoring the spec syntax to
path(<string>) | none

This has two major advantages:
- consistency with the only shipping implementation
- consistency with offset-path

ewilligers:
This brings the spec, the web platform tests and the only
shipping implementation (Chrome) into consistency:

  none | path(<string>)

closes #320


dirkschulze:
@ewilligers This does not seem correct. For compatibility reasons, the `d` CSS property must support the string version as well (beside the basic shape version). Furthermore, I am unsure if we want to limit `d` to just `path()` or support all basic shapes (including circle, ellipse and rectangle).

ewilligers:
> For compatibility reasons

Which implementations support a string version of the presentation attribute?

Do you mean consistency with the attribute? Note we already have unitless distance attributes (width etc) that have units in their CSS presentation attributes. Attribute syntax and CSS syntax are different.

AmeliaBR:
@dirkschulze 

For reference, there is discussion in the following issues:

- https://github.com/w3c/svgwg/issues/49
- https://github.com/w3c/svgwg/issues/119
- https://github.com/w3c/svgwg/issues/320

I still strongly disagree with using `path()` in any place other than as an interoperable `<shape-function>` where it can be substituted for any of the other shape functions.

But none of the other implementers have weighed in to argue for that approach, and Chrome has been shipping it following Eric's proposal.

I will point out that the current state (`path()` supported in some CSS properties, other shapes in other properties) has proven to be very confusing to authors: https://css-tricks.com/basic-shapes-path-never-twain-shall-meet/

The proposal should probably be discussed with CSS WG as well as SVG.

dirkschulze:
@ewilligers The change you made affects the `d` property _as well_ as the `d` presentation attribute. So you pretty much disallow path strings everywhere. This is the compatibility issue I mean. Furthermore, it is confusing that one will be allowed to use path strings on the `d` presentation attribute but not on the `d` CSS property. You can't simply copy the content from the attribute to the style section and are done with it. You must add the `path()` function. IMO there is no benefit to not allow the path string even if it is for the `d` property only. We are not getting away with the path string on the `d` presentation attribute ever. So why forcing its removal on the CSS property?

Also, the spec text should say what happens when a `<fill-rule>` was specified. How does this interact with the `fill-rule` CSS property? CSS Masking has a similar problem with `clip-rule` and `clip-path`.

@AmeliaBR I totally agree with you. IMO it is fine that Chrome starts shipping with `path()` and adds support for other shapes later but spec wise we should not split the shape functions.

IMO we should not make this change for SVG2 at all and instead split the `d` property out in general into its own spec or another existing FXTF/CSSWG spec. The path data string logic can stay in SVG2 for now.

fsoder:
I'd agree with @dirkschulze that there would need to be additional _prose_ for the presentation attribute form of `d` - because regardless of `<string>` or `<path()>` you cannot just pass the attribute value to the CSS parser and get something path'y back (a jumble of tokens is what you'll be getting.) (An attribute value - string - isn't a `<string-token>`. You need quotes and potentially escaping to produce the latter from the former.) So since pre-processing is going to be required _anyway_, one might as well go with what seems like the lesser legacy form (`<path()>`), and avoid adding another (`<string>`).

dirkschulze:
@ewilligers Do we want to move this change to SVG 2.1 or rather level 1 of a path spec?

ewilligers:
I have updated the pull request to describe the difference between the property and attribute syntax.

master/types.html
```diff
@@ -144,7 +144,11 @@ <h2 id="syntax">Attribute syntax</h2>
 unitless length and angles to be used in presentation attribute while
 disallowing them in corresponding property values.</p>
 
-<p class="note">Note that all <a>presentation attributes</a>, since they are
+<p>The <a>'d'</a> presentation attribute is an exception, it is parsed
```
dirkschulze: For CSS Transforms we are a little bit more verbose and explain why there are differences as well. See https://drafts.csswg.org/css-transforms/#svg-syntax I think we should be a little bit more verbose here too.

Did we agree on freezing the `d` presentation attribute to the definition of SVG 1.1? No support of the additional CSS property values?

IMO we should allow all values the property has + the string value itself for backwards compatibility reasons. But I am not sure if:
1. We have discussed this in the WG already and resolved to not extend the grammar
2. Chrome would support the additional CSS syntax on the attribute as well.

dirkschulze:
@ewilligers I’d love to see this in SVG. At this point I wonder if we should defer it and look at adding this to the path spec as an extension to SVG2.0.

AmeliaBR:
@dirkschulze Based on the resolutions we made in August (https://github.com/w3c/svgwg/issues/320#issuecomment-414462645), the revised spec would match one implementation (Chrome). Considering how eagerly anticipated it is by authors, I don't think it's appropriate to remove just yet.

(But @ewilligers, any chance you could update that PR so we can land it?)

dirkschulze:
@AmeliaBR I'd love to keep it but we need 2 implementations for transitioning the spec. Other implementations might be willing to implement it but this needs to happen before the transition. Lets discuss how we want to deal with features that have implementers feedback but no 2 implementations yet in our next WG call.

ewilligers:
@AmeliaBR I have updated the PR with the attribute specification.

master/paths.html
```diff
@@ -231,6 +231,10 @@ <h3 id="TheDProperty">Specifying path data: the <span class='property'>'d'</span
 If the path data string contains no valid commands, then the behavior
 is the same as the <span class='prop-value'>none</span> value.</p>
 
+<p>When <a>'d'</a> is parsed as an attribute, it is parsed according to the
```
dirkschulze: Could you be more specific what this means for the path() function please? I thing what you want to say is that the grammar would not allow path() and falls back to EBNF as specified in this spec.

master/types.html
dirkschulze: Guess I commented here already. Still applies IMO.

master/paths.html
```diff
@@ -177,7 +177,7 @@ <h3 id="TheDProperty">Specifying path data: the <span class='property'>'d'</span
   </tr>
   <tr>
     <th>Value:</th>
-    <td>none | <a>&lt;string&gt;</a></td>
+    <td>none | path(<a>&lt;string&gt;</a>)</td>
```
dirkschulze: I wonder if we want to link `path()` (and maybe string as well).

master/paths.html
```diff
@@ -243,26 +247,51 @@ <h3 id="TheDProperty">Specifying path data: the <span class='property'>'d'</span
...
   parameter to each path data command must be
   <a href="https://drafts.csswg.org/web-animations/#animation-interpolation">interpolated</a>
   separately <a href="https://drafts.csswg.org/web-animations/#real-number-animation-type">as
   real numbers</a>.  Flags and booleans must be interpolated as
-  fractions between zero and one, with any non-zero value considered
-  to be a value of one/true.
+  integer values clamped between zero and one.
 </p>
 
-<p class="annotation">
```
dirkschulze: We let the annotation in the spec for later reference.

dirkschulze:
`d` is listed in the presentation attribute list already. It should also be added to https://svgwg.org/svg2-draft/propidx.html

Please check if the changes.html section needs updates.

master/propidx.html
```diff
@@ -96,6 +96,16 @@ <h1>Property Index</h1>
...
         </tr>
+        <tr>
+          <th><a>'d'</a></th>
+          <td>none | path(<a>&lt;string&gt;</a>)</td>
+          <td>none</td>
+          <td><a>'path'</a> elements</td>
+          <td>no</td>
+          <td>N/A</td>
+          <td><a href="https://www.w3.org/TR/2008/REC-CSS2-20080411/media.html#visual-media-group">visual</a></td>
+          <td>yes</td>
```
AmeliaBR: We've updated the Property Index to specify the type of animation, not just yes/no. There's also a "computed value" column. Which makes me realize that we never resolved #321 on what that should be (do you normalize before computed/used value?).  For now:

```suggestion
          <td>as path data string</td>
          <td>as specified</td>
```

Ideally, "as path data string" would be linked to the definition of how animation works.

master/types.html
AmeliaBR: I think it would confuse parsing too much to to allow the `path()` notation inside the attribute.

That said, I don't see any reason to freeze the attribute at SVG 1.1 when it comes to path data commands. Instead, treat the attribute value as a string injected into a `path()` wrapper. If we ever do add other commands, they would be valid here, too.

svgeesus:
Agenda+ added so we can decide what to do with this rather old PR

svgeesus:
There is a minor conflict on propidx.html which looks very easy to rebase( the current version has some property deletions).

vwkd:
What's the progress on this? It's really a pity that the `d` path attribute doesn't have a working CSS property equivalent.

Can't the WG make an effort to resolve this PR that's hanging for now almost 4 years? It's funny when the last words on this is from before 2 1/2 years saying "I'll try to do the work this week"....

vwkd:
Can the WG make an effort to resolve this PR thats hanging now for almost 5 years?

Issue #385, opened 2018-02-05
Review implementation support - Paths chapter


ewilligers:
[Changes:](https://svgwg.org/svg2-draft/changes.html#paths)

### Added new B and b "bearing" path commands. [At Risk]
[Tests](https://wpt.fyi/svg/path/bearing) show no browsers support bearing commands.

### Allow Z or z to fill in missing path coordinate data in the previous command with the coordinate of the initial point in the subpath. [At Risk]
[Tests](https://wpt.fyi/svg/path/closepath) show no browsers support z filling in missing path coordinate data.

### Define ‘path’, ‘polygon’ and ‘polyline’ elements with no data set (empty or zero valid commands) to not render.
[Tests](https://wpt.fyi/svg/path/error-handling) show Chrome, Edge, Firefox, Safari support this functionality.

### Removed the SVGPathSeg* and SVGAnimatedPathData interfaces and the related methods on SVGPathElement.
[Tests](https://wpt.fyi/svg/path/interfaces/SVGAnimatedPathData-removed.svg) show only Chrome has removed the interfaces and related methods. 

### Promoted the ‘d’ attribute to a property. [At Risk]
[Tests](https://wpt.fyi/svg/path/property) show Chrome exposes 'd' as a property, with an implementation that [differs](https://github.com/w3c/svgwg/pull/374) from the current spec.

### Removed the pathLength attribute, getTotalLength(), and getPointAtLength() methods from SVGPathElement, they are now on SVGGeometryElement. [At Risk]
[Tests](https://wpt.fyi/svg/types/elements/SVGGeometryElement-rect.svg) show lack of browser support.

### Clarified that a value of zero for ‘pathLength’ is valid.
[Tests](https://wpt.fyi/svg/path/distance) exist. There are some test failures, these need investigation.

longsonr:
Firefox nightlies support pathLength attribute, getTotalLength(), and getPointAtLength() on SVGGeometryElement. All shapes are now SVGGeometryElement and Firefox nightly passes the SVGGeometryElement-rect.svg test.

Firefox is also removing the parts of the SVGPathSeg* and SVGAnimatedPathData interfaces that manipulate paths. The interface remains currently but only allows retrieval of path data.

boggydigital:
Not blocking updated 2.0 CR publication - assigning 2.0 Recommendation milestone to clean this up before 2.0 REC

Tavmjong:
[...]

PR #346, opened 2017-09-16
Fixed path grammar 

fuchsia:
As chronicled in #335, this extends the path grammar to allow decimals and exponents on numbers, as well as correcting all the defects described in this [comment](https://github.com/w3c/svgwg/issues/335#issuecomment-320349171). 

The EBNF used is that laid out for [XML](https://www.w3.org/TR/REC-xml/#sec-notation). It has been tested with a homebrew script that validates strings against EBNF.

PR #398, opened 2018-03-30
Remove support for segment-completing close path

ewilligers:
Segment-completing close path commands, where coordinates are omitted
and implicitly replaced with the initial point of the subpath,
lack implementations and have been removed from the spec.

Discussed in #385


ewilligers:
The grammar changes become redundant if https://github.com/w3c/svgwg/pull/346 proceeds. That PR can be reviewed first, then I'll rebase so only my prose edits remain.

dirkschulze:
@liamquin There seem to be questions about moving this to the path spec?

dirkschulze:
Thought the WG logs seem to have a limited context to this question. CC @boggydigital

ewilligers:
> There seem to be questions about moving this to the path spec?

It remains in the path spec (like bearing and Catmull Rom commands), we are simply removing it from SVG 2.

This is ready for review.

master/paths.html
```diff
@@ -340,33 +340,13 @@ <h3 id="PathDataMovetoCommands">The <strong>"moveto"</strong> commands</h3>
 <h3 id="PathDataClosePathCommand">The <strong>"closepath"</strong> command</h3>
 
 <p>The "closepath" (<strong>Z</strong> or <strong>z</strong>)
-  must end the current subpath by connecting it back to its <a>initial point</a>
-  in either of two ways:
+  ends the current subpath by connecting it back to its <a>initial point</a>.
+  An automatic
+  straight line is drawn from the current point to the <a>initial point</a>
+  of the current subpath. This <a>path segment</a> may be of zero
+  length.
```
dirkschulze: The original text also said what the next initial point is when "closepath" is not followed by a moveto.

dirkschulze: Also, IMO it is important to keep the original sentence that a lineto to the initial point is not the same as specifying a closepath.

Maybe we should copy those sentences over: https://www.w3.org/TR/SVG11/paths.html#PathDataClosePathCommand

ewilligers: The sentences have now been copied.

ewilligers:
Please review. The statements from SVG 1.1 have been restored.

rolivo:
Jesus holy, there really is some difference between SVG 1.1 and SVG 2.0, if you have removed all the new features that you had.

Issue #315, opened 2017-05-07
SVGAElement all attributes except href should be as HTML i.e. not SVGAnimatedString


longsonr:
download, rel, relList, hreflang and type are all modelled as SVGAnimatedString attributes. There seems no reason to do this as they aren't much use as things to be SMIL animated (in fact download is stated to be non-animatable in the spec).

  Possible solution

    have these the same as the HTML specification i.e. 

           [SetterThrows]
           attribute DOMString download;
           [SetterThrows]
           attribute DOMString rel;
           [PutForwards=value]
           readonly attribute DOMTokenList relList;
           [SetterThrows]
           attribute DOMString hreflang;
           [SetterThrows]
           attribute DOMString type;

AmeliaBR:
Closing as duplicate of #312. @longsonr, could you please add your suggestions to that issue?

AmeliaBR:
Hmm... unless you want to make this a separate issue, just for the second point.  In which case, edit the title accordingly & I'll re-open.

For reference: I made them animated strings for consistency with the SVG 1.1 properties on the interface.  But I agree that there's no real use for them.

longsonr:
Let's make it a separate issue.

longsonr:
@AmeliaBR I've narrowed it to just the non-href attributes. Can you reopen it please?

dstorey:
I agree this is a good idea, especially as they're not currently implemented (AFAICT) with the exception of download in Firefox, but [they implemented it as DOMString anyway](https://dxr.mozilla.org/mozilla-central/source/dom/webidl/SVGAElement.webidl).

We could even possibly look to pull them in from HTML as part of a mixin if they matched, which would keep them in sync.

liamquin:
@longsonr , can Mozilla contribute any tests in this area to wpt for svg, assuming we change them to DOMString?

liamquin:
[either way, we need to mark Dom support for these attributes as at risk in SVG unless we know we're getting 2 implementations this Summer.]

dstorey:
I agree with @dirkschulze about avoiding any new SVGAnimatedString.  We (Edge) are unlikely to implement these in the current form. Sharing the definition/implementation with HTML would make it more likely that we would implement.

dstorey:
There are presumably 3 implementers (@dirkschulze, @longsonr, myself) in favour of using the HTML definition rater than SVGAnimatedString. 

[...]

PR #409, opened 2018-04-10
Update SVGAElement to match attributes on HTMLAnchorElement

dstorey:
* Harmonised existing IDL attributes with the definitions in HTML (inc. setting Animatable to false and no longer read only)
* Added IDL attributes missing from SVG but included in HTML - ping, referrerPolicy and text. These 3 are in the process of being implemented in Firefox.

This is temporary until the HTML spec includes the HTMLOrSVGAnchorElement mixin (or equivalent), when we can just refer to that spec. I just copied the reflect rules for `text` from HTML for example. 

Apologies if I didn't add things in all the right places or got the references wrong. I'm still learning how things should be structured and edited.


dstorey:
Tests for some of these: https://github.com/w3c/web-platform-tests/pull/10275
Patch for Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1451823

master/definitions.xml
```diff
@@ -16,11 +16,13 @@
       interfaces='SVGAElement'>
     <x:contentmodel xmlns='http://www.w3.org/1999/xhtml'>Descriptive content, plus any element or text allowed by its parent's content model, except for another a element.  If the parent is a switch element, use the content model of the nearest ancestor that isn't a switch.</x:contentmodel>
     <attribute name='href' href='linking.html#AElementHrefAttribute' animatable='yes'/>
-    <attribute name='target' href='linking.html#AElementTargetAttribute' animatable='yes'/>
-    <attribute name='download' href='linking.html#AElementDownloadAttribute' animatable='yes'/>
-    <attribute name='rel' href='linking.html#AElementRelAttribute' animatable='yes'/>
-    <attribute name='hreflang' href='linking.html#AElementHreflangAttribute' animatable='yes'/>
-    <attribute name='type' href='linking.html#AElementTypeAttribute' animatable='yes'/>
+    <attribute name='target' href='linking.html#AElementTargetAttribute' animatable='no'/>
```
dirkschulze: `target` is animatable in WebKit as far as I can see. This was implemented before the Blink split so I am positive that it is in Blink as well. Could you add the link to the WG resolution if this was resolved on please?

ewilligers: target is an SVGAnimatedString in SVG 1.1 and in all browsers including Edge.

The following example works in all browsers, using target.baseVal to read the target.
https://jsfiddle.net/ericwilligers/bL9zvd1w/

dstorey: we agreed to change target to DOMString as the usage is so tiny.

AmeliaBR: Here's the resolution for that decision:
https://github.com/w3c/svgwg/issues/315#issuecomment-376271318

AmeliaBR:
Okay, I'm merging in the changes to everything except `target`, as discussed on the call today (see https://github.com/w3c/svgwg/pull/409#issuecomment-412654080).

This means the spec matches the Firefox implementation.  And hopefully means that other browsers can move ahead with matching _their_ implementations!

We'll continue discussion of the potential breaking change re `target` in a separate issue (which I'll link here once it exists).

PR #413, opened 2018-04-12
Add DocumentAndElementEventHandlers

### LINKED ISSUE #395: Should we move more HTMLElement members to HTMLorSVGElement mixin?
@annevk made the good point when [making the change to include a HTMLorSVGElement mixin](https://github.com/whatwg/html/pull/3543#issuecomment-375579673) that there are other members on HTMLElement that would probably make sense to include in SVGElement via the mixin. click() is one example.

I'd like us to discuss on the call if we want to adopt any of these for SVG. For reference:
* [newly defined HTMLOrSVGElement mixin
](https://html.spec.whatwg.org/multipage/dom.html#htmlorsvgelement)
* [The other members currently on HTMLElement](https://html.spec.whatwg.org/multipage/dom.html#htmlelement)

dstorey:
Fixes #395  (any additional members should have their own issue)

Supported by Firefox and Edge (upcoming version) as specified. Supported by Chrome and Safari on Element (HTML defines them on HTMLElement so Safari/Chrome have them one level too high)

Tests: https://github.com/w3c/web-platform-tests/pull/10249

### DIFF
--- a/master/changes.html
+++ b/master/changes.html
@@ -130,7 +130,7 @@ have been made.</p>
 
   <li>Added <a href="struct.html#__svg__SVGDocument__activeElement">activeElement</a> attribute to <a>Document</a>.</li>
 
-  <li>Made <a>SVGElement</a> implement the <a>GlobalEventHandlers</a> interface from HTML.</li>
+  <li>Made <a>SVGElement</a> include the <a>GlobalEventHandlers</a> interface from HTML.</li>
 
   <li>Removed getStrokeBBox from <a>SVGGraphicsElement</a> and extended <a href="types.html#__svg__SVGGraphicsElement__getBBox">getBBox</a> with a dictionary argument that controls which parts of the element are included in the returned bounding box.</li>
 
@@ -178,6 +178,7 @@ have been made.</p>
   <li><a>SVGUnitTypes</a> is no longer NoInterfaceObject. <a href="https://github.com/w3c/svgwg/issues/291">Github #291</a>.</li>
   <li>Clarified that <a>SVGElement</a>.<a href="types.html#__svg__SVGElement__className">className</a>
       overrides <a>Element</a>.className. Not normative, <a href="https://github.com/w3c/svgwg/issues/298">Github #298</a>.</li>
+  <li>Made <a>SVGElement</a> include the <a>DocumentAndElementEventHandlers</a> interface from HTML.</li>
 </ul>
 </div>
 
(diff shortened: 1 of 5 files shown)


dstorey:
Do I need to make something like:
```
<attributecategory
      name='document and element event'
```
that lists all event handler attributes in DocumentAndElementEventHandlers and add them to all 44 places in SVG? This seems to defy the point in mixins if we have to maintain every content attribute that reflects a IDL attribute in a mixin that is defined in another spec.

AmeliaBR:
> Do I need to make something ... that lists all event handler attributes

Good point. Technically, the mixin only defines the DOM interface properties. It doesn't add the attributes to the element model. So the attributes do need to be defined separately.

That said, the text in the [Event Handlers section](https://svgwg.org/svg2-draft/interact.html#EventAttributes) in the interactivity chapter is pretty open-ended, automatically including any new event types:

> event attribute
> An event attribute always has a name that starts with "on" followed by the name of the event for which it is intended. It specifies some script to run when the event of the given type is dispatched to the element on which the attribute is specified.
> 
> For every event type that the user agent supports, SVG supports that as an event attribute, following the same requirements as for event handler content attributes [HTML]. The event attributes are available on all SVG elements.

So maybe a better strategy would be to get rid of the list of explicit attributes for the "global event attributes" category, and instead make it link to that section.  But I'm not sure where in the build process we need to change something to make that happen.

dstorey:
Until we work out the build process, I'll just add DocumentAndElementEventHandlers  every where that GlobalEventHandlers is included for now. As it is for cut/copy/paste it probably won't change too much, unless onbeforecut/copy/paste are standardized (which are supported by all browsers)

master/changes.html
```diff
@@ -178,6 +178,7 @@ <h3 id="types">Basic Data Types and Interfaces chapter (owner: BogdanBrinza)</h3
   <li><a>SVGUnitTypes</a> is no longer NoInterfaceObject. <a href="https://github.com/w3c/svgwg/issues/291">Github #291</a>.</li>
   <li>Clarified that <a>SVGElement</a>.<a href="types.html#__svg__SVGElement__className">className</a>
       overrides <a>Element</a>.className. Not normative, <a href="https://github.com/w3c/svgwg/issues/298">Github #298</a>.</li>
+  <li>Made <a>SVGElement</a> include the <a>DocumentAndElementEventHandlers</a> interface from HTML.</li>
```
dirkschulze: Doesn't this require a "changed since last publication" div? Same for the above?

dstorey: this already is in the changed-since-cr1 or has there been another CR and this and the previous change needs to go into a changed-since-cr2 div?

dstorey:
@dirkschulze / @AmeliaBR : this should be ready for final review I believe.

Issue #298, opened 2016-12-05
Add advice on how to handle className conflict in DOM?


BigBadaboom:
Re: Interface SVGElement
https://svgwg.org/svg2-draft/single-page.html#types-InterfaceSVGElement

As far as I can tell, this is the first version of the spec since className was added to the Element interface in DOM level 2 (2002).  SVGElement in SVG 1.2 Tiny didn't include className in its uDOM definition.

Should the spec provide guidance on how implementations should handle the conflict between SVGElement.className and Element.className?

AmeliaBR:
What I'd *like* to do is to have SVGElement.className act as a regular DOM string that has special properties `.animVal` and `.baseVal` that implement the old SVGAnimatedString behavior.  See https://github.com/w3c/svgwg/issues/175 for my musings about speccing that sort of behavior SVG-wide.

In the short term, however, based on existing implementations, the guidance is more that "the `.className` property of SVGElement *overrides* the property of the same name on the `Element` interface".

Of course, whether this is easy or hard to implement depends on the type-strictness of the language in which you're implementing the DOM.

nikosandronikos:
Since `SVGElement.className` is deprecated. It would be nice if we could reverse the order of that and defer to `Element.className` if it exists.

AmeliaBR:
> Since SVGElement.className is deprecated. It would be nice if we could reverse the order of that and defer to Element.className if it exists.

That wouldn't be backwards compatible, and doesn't get at the main issue: should implementers move their existing HTMLElement.classname implementations to be part of the core Element interface, or no?  If we're saying they should, then SVGElement.classname isn't just deprecated, it's obsolete & any old SVG script references to it will be broken.

AmeliaBR:
On the [call today](http://www.w3.org/2016/12/22-svg-minutes.html) we agreed that--as much as we'd like a more compatible option--for now, we want to spec the current behavior.  Meaning that the SVGElement definition of `className` would override the inherited definition from `Element`.

@nikosandronikos has agreed to look into the [WebIDL rules on overloading inherited properties](https://heycam.github.io/webidl/#idl-overloading) to make sure that's allowed, and to add a short clarification/warning to the spec.

nikosandronikos:
The important text in [WebIDL](https://www.w3.org/TR/WebIDL/#dfn-inherit) is:

[...]

PR #415, opened 2018-04-18
Add HTMLOrSVGElement mixn

### LINKED ISSUE #60: Move dataset & other shared IDL attributes from SVGElement to HTMLOrSVGElement
The current draft of the SVG spec defines `dataset` attribute on the [SVGElement](https://svgwg.org/svg2-draft/single-page.html#types-InterfaceSVGElement) interface while HTML 5 spec defines it on the [HTMLElement](https://html.spec.whatwg.org/multipage/dom.html#elements-in-the-dom) interface.

Is there any particular reason why `dataset` couldn't be defined just once on the `Element` interface like other similar attributes such as `classList`?

dstorey:
Fixes #60 
Remove definiion of dataset/focus/blur/tabIndex and include mixin from HTML instead.

I've kept the content attribute definition of tabindex until we figure out what to do with the reflecting content attributes that pair with IDL attributes in mixins.

### DIFF
--- a/master/changes.html
+++ b/master/changes.html
@@ -178,6 +178,7 @@ have been made.</p>
   <li><a>SVGUnitTypes</a> is no longer NoInterfaceObject. <a href="https://github.com/w3c/svgwg/issues/291">Github #291</a>.</li>
   <li>Clarified that <a>SVGElement</a>.<a href="types.html#__svg__SVGElement__className">className</a>
       overrides <a>Element</a>.className. Not normative, <a href="https://github.com/w3c/svgwg/issues/298">Github #298</a>.</li>
+  <li>Moved dataset, tabIndex, focus(), and blur() to a shared <a>HTMLOrSVGElement</a> mixin defined in the HTML specification.</li>
 </ul>
 </div>
 
--- a/master/types.html
+++ b/master/types.html
@@ -1584,8 +1584,8 @@ or a property.  The way this reflection is done depends on the type of the IDL a
 <ul>
   <li>If the type of the reflecting IDL attribute is a
   <a href="http://heycam.github.io/webidl/#dfn-primitive-type">primitive type</a> (such as
-  <b>long</b>, as used by tabIndex 
-  on <a>SVGElement</a>) or <b>DOMString</b> (as used by
+  <b>long</b>, as used by <a href="https://html.spec.whatwg.org/multipage/interaction.html#dom-tabindex">tabIndex</a> 
+   on <a>SVGElement</a>) or <b>DOMString</b> (as used by
   <a href="styling.html#__svg__SVGStyleElement__title">title</a> on <a>SVGStyleElement</a>),
   then the rules for
   <a href="https://html.spec.whatwg.org/multipage/infrastructure.html#reflect">reflecting
--- a/master/types.html
+++ b/master/types.html
@@ -225,18 +225,13 @@ interface <b>SVGElement</b> : <a>Element</a> {
 
   [SameObject] readonly attribute <a>SVGAnimatedString</a> <a href="types.html#__svg__SVGElement__className">className</a>;
 
-  [SameObject] readonly attribute <a>DOMStringMap</a> <a href="types.html#__svg__SVGElement__dataset">dataset</a>;
-
   readonly attribute <a>SVGSVGElement</a>? <a href="types.html#__svg__SVGElement__ownerSVGElement">ownerSVGElement</a>;
   readonly attribute <a>SVGElement</a>? <a href="types.html#__svg__SVGElement__viewportElement">viewportElement</a>;
-
-  attribute long <a href="types.html#__svg__SVGElement__tabIndex">tabIndex</a>;
-  void <a href="types.html#__svg__SVGElement__focus">focus</a>();
-  void <a href="types.html#__svg__SVGElement__blur">blur</a>();
 };
 
 <a>SVGElement</a> includes <a>GlobalEventHandlers</a>;
-<a>SVGElement</a> includes <a>SVGElementInstance</a>;</pre>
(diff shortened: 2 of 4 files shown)


master/types.html
```diff
@@ -225,18 +225,13 @@ <h3 id="InterfaceSVGElement">Interface SVGElement</h3>
...
-
-  attribute long <a href="types.html#__svg__SVGElement__tabIndex">tabIndex</a>;
-  void <a href="types.html#__svg__SVGElement__focus">focus</a>();
-  void <a href="types.html#__svg__SVGElement__blur">blur</a>();
 };
 
 <a>SVGElement</a> includes <a>GlobalEventHandlers</a>;
-<a>SVGElement</a> includes <a>SVGElementInstance</a>;</pre>
+<a>SVGElement</a> includes <a>SVGElementInstance</a>;
+<a>SVGElement</a> includes <a>HTMLOrSVGElement</a></pre>
```
dirkschulze: Semicolon seems to missing at last line.

master/types.html
```diff
@@ -1614,8 +1584,8 @@ <h2 id="DOMInterfacesForReflectingSVGAttributes">DOM interfaces for reflecting S
 <ul>
   <li>If the type of the reflecting IDL attribute is a
   <a href="http://heycam.github.io/webidl/#dfn-primitive-type">primitive type</a> (such as
-  <b>long</b>, as used by <a href="types.html#__svg__SVGElement__tabIndex">tabIndex</a>
-  on <a>SVGElement</a>) or <b>DOMString</b> (as used by
+  <b>long</b>, as used by <a href="types.html#__svg__SVGNameList__numberOfItems">numberOfItems</a>
+  on <a>SVGNumberList</a>) or <b>DOMString</b> (as used by
```
dirkschulze: This seems to be an additional change that doesn't seem to belong to the mixin change. I am unsure if this general reference to a `SVGNumberList` item is correct (also why not simply referencing `SVGNumberList` instead of giving an obscure example?)

Also, SVGNumberList uses `unsigned long` and not `long`.

dstorey: I was removing the tabIndex reference as it is no longer defined in the spec and thus the link wouldn't work. I originally went with baseVal as that is a long in the same section but it was confusing with the section below it that talks about SVGAnimatedString. I guess I could find another example. (SVGNumberList links to that generic SVGNameList as many share the same definition, so it was probably a bad example to pick)

dstorey:
yeah the content attribute is still in but the link in the change above is for the IDL attribute. I'll look into adding that definition to HTML thanks!

dstorey:
I couldn't find where IDL attributes are defined in definitions. `<attribute>` is content attributes. The things defined in other specs as far as I could see are terms, css properties and interfaces. Is there a special element type I should be using for IDL attributes in there?

dirkschulze:
@tobie do you help with the question from @dstorey ?

tobie:
I’m not sure I understand the question. Is this about what kind of markup should be used to reference an attribute?

dstorey:
Yeah, the original markup was `<a href="types.html#__svg__SVGElement__tabIndex">tabIndex</a>`, but now that tabIndex IDL attribute is defined in HTML that ID wouldn't be fond. Amelia suggested adding tabIndex to definitions.xml linking to the HTML spec, so that <a>tabIndex</a> would like correctly (if I understood correctly). But I don't see how to add IDL attributes (rather than content attributes or interfaces) to definitions.xml.

tobie:
Oh! I'm the wrong person to ask this. You need to talk to however wrote the system you're using to write the spec.

AmeliaBR:
> But I don't see how to add IDL attributes (rather than content attributes or interfaces) to definitions.xml.

Ah, sorry for the confusion. Hadn't actually looked at it, didn't realize that the auto-linking only included complete interfaces, and not interface members.

If there is only one `<a>` element that links to `tabIndex`, you can hard-code a link to https://html.spec.whatwg.org/multipage/interaction.html#dom-tabindex

dirkschulze:
@tobie Sorry! Misunderstood the issue from @dstorey :P

dstorey:
@dirkschulze / @AmeliaBR I think this one is ready now?

master/changes.html
```diff
@@ -178,6 +178,7 @@ <h3 id="types">Basic Data Types and Interfaces chapter (owner: BogdanBrinza)</h3
   <li><a>SVGUnitTypes</a> is no longer NoInterfaceObject. <a href="https://github.com/w3c/svgwg/issues/291">Github #291</a>.</li>
   <li>Clarified that <a>SVGElement</a>.<a href="types.html#__svg__SVGElement__className">className</a>
       overrides <a>Element</a>.className. Not normative, <a href="https://github.com/w3c/svgwg/issues/298">Github #298</a>.</li>
+  <li>Moved dataset, tabIndex, focus(), and blur() to a shared <a>HTMLOrSVGElement</a> mixin defined in the HTML specification.</li>
```
dirkschulze: oops, forgot to say that this requires a changed-since-last-cr div... see other things in the file.

dstorey: This is already in the changes since CR section:

`<div class='changed-since-cr1'>
<ul>
  <li><a>SVGUnitTypes</a> is no longer NoInterfaceObject. <a href="https://github.com/w3c/svgwg/issues/291">Github #291</a>.</li>
  <li>Clarified that <a>SVGElement</a>.<a href="types.html#__svg__SVGElement__className">className</a>
      overrides <a>Element</a>.className. Not normative, <a href="https://github.com/w3c/svgwg/issues/298">Github #298</a>.</li>
  <li>Moved dataset, tabIndex, focus(), and blur() to a shared <a>HTMLOrSVGElement</a> mixin defined in the HTML specification.</li>
</ul>
</div>`

PR #425, opened 2018-04-27
Add or update dfns for definitions

### LINKED ISSUE #417: Use <dfn> for definitions
Cross-linking to SVG terms is broken because y'all aren't using `<dfn>`.  E.g. Terms in https://www.w3.org/TR/SVG2/render.html#Definitions can't be linked to with Bikeshed.

Please mark up all your terms with `<dfn>`, even if they're in a `<dt>`. (Because `<dt>` isn't always a term being defined, we don't index `<dt>`; defining instances of a term need to use the defining instance tag, `<dfn>`.) Note you'll also need to move the ID from the `<dt>` to the `<dfn>` so that it's correctly associated.

dstorey:
Fixes #417 

I've only touched the main spec and not the various sub specs, as that will take a while.

I included data-dfn-exports (seems to be a typo of using dfn-exports in the spec) to most things. I don't know if this is correct, but it seems if we want to mark things up as definitions then they are thinks that should be useful to export/link to.

### DIFF
--- a/master/coords.html
+++ b/master/coords.html
@@ -21,16 +21,16 @@
 <div class="ready-for-wider-review">
 
 <p>All SVG content is drawn inside
-<dfn data-dfn-type="dfn" id="TermSVGViewport">SVG viewports</dfn>.
+<dfn id="TermSVGViewport" data-dfn-type="dfn" data-dfn-export>SVG viewports</dfn>.
 Every SVG viewport defines a drawing region characterized by a size 
 (width, height), and an origin, measured in abstract 
-<dfn data-dfn-type="dfn" id="TermUserUnits">user units</dfn>.</p>
+<dfn id="TermUserUnits" data-dfn-type="dfn" data-dfn-export>user units</dfn>.</p>
 
 <p class="note">Note that the term SVG viewport is distinct from the 
 <a href="https://www.w3.org/TR/2011/REC-CSS2-20110607/visuren.html#viewport">"viewport"</a>
 term used in CSS.</p> 
 
-<p>The <dfn data-dfn-type="dfn">initial viewport</dfn> is a top-level
+<p>The <dfn id="TermInitialViewport" data-dfn-type="dfn" data-dfn-export>initial viewport</dfn> is a top-level
 SVG viewport that establishes a mapping between the coordinate system used
 by the containing environment (for example, CSS pixels in web browsers)
 and <a>user units</a>. Establishing an initial viewport is described in more
@@ -41,8 +41,8 @@ detail in <a href="#ViewportSpace">The initial viewport</a>.</p>
 on which elements generate viewports.</p>
 
 <p>Each SVG viewport generates a 
-<dfn data-dfn-type="dfn" id="TermViewportCoordinateSystem">viewport coordinate system</dfn>
-and a <dfn data-dfn-type="dfn">local coordinate system</dfn>, initially identical.
+<dfn id="TermViewportCoordinateSystem" data-dfn-type="dfn" data-dfn-export>viewport coordinate system</dfn>
+and a <dfn id="TermLocalCoordinateSystem" data-dfn-type="dfn" data-dfn-export>local coordinate system</dfn>, initially identical.
 Providing a <a>'viewBox'</a> on a viewport's element transforms the local coordinate system 
 relative to the viewport coordinate system as described in
 The <a>'viewBox'</a> attribute. Child elements of a viewport can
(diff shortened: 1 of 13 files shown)


master/struct.html
```diff
@@ -2302,7 +2301,7 @@ <h3 id="roleattribute">Role attribute</h3>
         <th>Animatable</th>
       </tr>
       <tr>
-        <td><dfn id="RoleAttribute">role</dfn></td>
+        <td><dfn id="RoleAttribute" data-dfn-type="element-attr" data-dfn-export>role</dfn></td>
```
dstorey: this needs a data-dfn-for, but I've no idea how to say apply to all elements without listing every single SVG element.

master/types.html
```diff
@@ -1271,7 +1271,7 @@ <h3 id="ListInterfaces">List interfaces</h3>
   </li>
 </ol>
 
-<p>Whenever a list element object is to be <dfn id="TermDetach">detached</dfn>,
+<p>Whenever a list element object is to be <dfn id="TermDetach" data-dfn-type="dfn" data-dfn-export>detached</dfn>,
```
dstorey: There are some like these that are very generic words. Not sure if they should be exported or have more descriptive labels applied.

master/types.html
```diff
@@ -1578,7 +1578,7 @@ <h2 id="DOMInterfacesForReflectingSVGAttributes">DOM interfaces for reflecting S
 
 </div>
 
-<p>Some IDL attributes <dfn id="TermReflect">reflect</dfn> the value of a content attribute
+<p>Some IDL attributes <dfn id="TermReflect" data-dfn-type="dfn">reflect</dfn> the value of a content attribute
```
dstorey: I figured reflect must be cononically defined in some other spec as it is a common DOM/HTML term. so didn't add the data-dfn-exports. Should these type of terms be marked up as dfn or just linked to whichever of DOM or HTML that defines it? Same applies to a few other things in SVG like stacking contexts etc.

dstorey:
This probably needs experts from each section to check to make sure I marked things up correctly.

master/types.html
```diff
@@ -1623,7 +1623,7 @@ <h2 id="DOMInterfacesForReflectingSVGAttributes">DOM interfaces for reflecting S
 </ol>
 
 <p>When a <a>reflected</a> content attribute is to be
-<dfn id="TermReserialize">reserialized</dfn>, optionally using a
+<dfn id="TermReserialize" data-dfn-type="dfn" data-dfn-export>reserialized</dfn>, optionally using a
```
dirkschulze: Is that a term we want to get referenced from outside? Seems very generic. But of course, another spec could always clarify which spec it wants to link to in case the term exists already.

master/coords.html
```diff
@@ -834,15 +834,15 @@ <h2 id="BoundingBoxes">Bounding boxes</h2>
...
+  <li>The <dfn id="TermObjectBoundingBox" data-dfn-type="dfn" data-dfn-export>object bounding box</dfn> is the bounding box that contains only
   an element's geometric shape.  For <a>basic shapes</a>, this is the area
   that is filled.  Unless otherwise specified, this is what is meant by the
   unqualified term "bounding box".</li>
 
-  <li>The <dfn id="TermStrokeBoundingBox" data-dfn-type="dfn" data-export="">stroke bounding box</dfn> is the bounding box that contains
+  <li>The <dfn id="TermStrokeBoundingBox" data-dfn-type="dfn" data-dfn-export>stroke bounding box</dfn> is the bounding box that contains
   an element's geometric shape and its <a>stroke shape</a>.</li>
 
-  <li>The <dfn id="TermDecoratedBoundingBox" data-dfn-type="dfn" data-export="">decorated bounding box</dfn> is the bounding box that contains
```
dirkschulze: Was the `data-export` set manually? On that note... our XML based tools are not affected by attributes w/o values?

master/interact.html
```diff
@@ -358,9 +358,9 @@ <h2 id="PointerEvents">Pointer events</h2>
 
 <h2 id="pointer-processing">Hit-testing and processing order for user interface events</h2>
 
-<dl class='definitions'  data-dfn-type="dfn" data-export="">
+<dl class="definitions">
```
dirkschulze: IIRC, `data-dfn-type` specified on a `dl` marks child `dt`s to the same type. You could check the spec.json that bikeshed downloads and see if those items show up as `dfn`.

master/coords.html
dstorey: I'm not sure but it was in a bunch of places. I'm not familiar with the XML tools, but there was a definitions class all over the source in single quotes. Wouldn't that break the XML tools too?

master/interact.html
dstorey: wasn't the original issue that these DLs didn't have dfn? The one called out was the DL on https://svgwg.org/svg2-draft/render.html#Definitions which has this pattern. But maybe it was due to using data-exports rather than data-dfn-exports?

dirkschulze: Maybe. Don't know.

dirkschulze:
I start seeing reference issues on CSSTransforms now: https://drafts.csswg.org/bikeshed/css-transforms-1/

Those likely are a result of this PR.

For example, "presentation attribute":
> No 'dfn' refs found for 'presentation attributes' that are marked for export.

https://github.com/w3c/svgwg/commit/cd0098490920dca08da61fec055a9355d5462e6c#diff-7165aa55468cbc8593fcbb8e9b5cb2ecR307

@plinss Is there a mistake that we did here? It seems like the dfn is marked for export.

PR #429, opened 2018-05-01
Fix lint/build issues

dstorey:
The multi-page versions of the SVG spec have gone from https://svgwg.org/svg2-draft/ . It looks from the linting emails that it was probably due to the missing attribute values (weirdly other xml failures like missing end tags and single quotes don't seem to phase it)

I've fixed those issues and the duplicate IDs in this PR

We should probably make the linter run on the GitHub side so that it warns you instead of giving you a green check mark, if anyone knows how to do that.

### DIFF
--- a/master/coords.html
+++ b/master/coords.html
@@ -21,16 +21,16 @@
 <div class="ready-for-wider-review">
 
 <p>All SVG content is drawn inside
-<dfn id="TermSVGViewport" data-dfn-type="dfn" data-dfn-export>SVG viewports</dfn>.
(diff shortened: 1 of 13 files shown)


dstorey:
@AmeliaBR / @dirkschulze this is quite important to get in if the multipage specs disappearing is related to the previous checkins, so if one of you could approve that would be great.

AmeliaBR:
PS. If you want to look into setting up pre-merge build validation, @dstorey, the person to ask for guidance would be @heycam, who set up much of the system initially.

PR #430, opened 2018-05-02
Fixes #405 update description of when visibility applies

### LINKED ISSUE #405: Wrong description of "visibility" property
[Chapter 3](https://svgwg.org/svg2-draft/render.html#VisibilityControl) says that:

> The visibility property only directly applies to graphics elements, text content elements (including the ‘a’ element when it is a child of text content element) and ‘use’ elements.

[Chapter 5](https://svgwg.org/svg2-draft/struct.html#UseStyleInheritance) says that:

> The visibility property does not appy directly to container elements, and therefore does not apply directly to the ‘use’ element.

Considering the recent changes to `<use>` element definition (make it behave like a container with referenced elements cloned into the shadow root), I guess the first sentence is wrong? Also, `<a>` is not a text content element, so I would avoid using the word "including". Therefore, I would propose to change the sentence from Chapter 3 to:

> The visibility property only directly applies to graphics elements, text content elements and "a" element (when it is a child of a text content element).

dstorey:


dstorey:
@dirkschulze / @AmeliaBR I think this is ready now?

AmeliaBR:
@dirkschulze That's a valid point. Maybe "directly affects rendering of" instead of "directly applies to"

PR #431, opened 2018-05-02
Fix #403 update opacity-value to alpha-value

### LINKED ISSUE #403: Replace <opacity-value> with <alpha-value>
[Property Index](https://svgwg.org/svg2-draft/propidx.html) makes use of `<opacity-value>` type which used to be defined in SVG 1.1 but is no longer defined in SVG 2 draft. I would propose to replace it with `<number> | <percentage>` type or `<alpha-value>` type which is defined by [CSS Color Module Level 4](https://www.w3.org/TR/css-color-4/#rgb-functions).

dstorey:


dstorey:
I didn't add anything in the change log as it is non-normative and propdix isn't included AFAICT

PR #432, opened 2018-05-02
Fix #402 Use CCS Colour L4 definition for computed value of *-opacity

### LINKED ISSUE #402: The computed value of fill-opacity stroke-opacity should never be a percentage
Currently the spec allows the computed value of [fill-opacity](https://svgwg.org/svg2-draft/painting.html#FillOpacityProperty) and [stroke-opacity](https://svgwg.org/svg2-draft/painting.html#StrokeOpacityProperty) to be either a number or a percentage.

This doesn't seem to be useful and it's inconsistent with the definition of "opacity" in [CSS Color Module Level 4](https://www.w3.org/TR/css-color-4/#transparency).

dstorey:


dstorey:
Notes:

* none of the CSS properties in pservers include a row for computed value, so I didn't change there. I just used the data type rather than number | percentage
* solid-opacity is not included in propidx
* flood-opacity is defined in the Filters spec. @dirkschulze: should that be updated to use alpha-value? it uses https://drafts.csswg.org/css-values-4/#typedef-number-percentage which is also number | percentage but is more general purpose (doesn't mention useful range being between 0 and 1)
* doesn't this need a change log entry or tests? Feels more like tightening up to expected platform conventions.

dirkschulze:
@dstorey It would be great if you could open an issue on https://github.com/w3c/fxtf-drafts/issues so that I won't forget about it.

PR #434, opened 2018-05-04
Update <length> | <percentage> to <length-percentage>

### LINKED ISSUE #190: Update percentage types to the new convenience notations
SVG 2 uses a lot of [`length | percentage`], this can be updated to `length-percentage` as defined in CSS Values and Units Level 3 ED once that spec reaches CR.
https://drafts.csswg.org/css-values-3/#percentages

dstorey:
Fixes #190.
Adds some minor reference clean up for number to remove hardcoded CSS2 references (number from CSS3 values is already in definitions.xml)

### DIFF
--- a/master/geometry.html
+++ b/master/geometry.html
@@ -32,7 +32,7 @@ describe the position and dimension of the <a>graphics element</a>s <a>'circle'<
   </tr>
   <tr>
     <th>Value:</th>
-    <td><a>&lt;length&gt;</a> | <a>&lt;percentage&gt;</a></td>
+    <td><a>&lt;length-percentage&gt;</a></td>
   </tr>
   <tr>
     <th>Initial:</th>
@@ -78,7 +78,7 @@ property</h2>
   </tr>
   <tr>
     <th>Value:</th>
-    <td><a>&lt;length&gt;</a> | <a>&lt;percentage&gt;</a></td>
+    <td><a>&lt;length-percentage&gt;</a></td>
   </tr>
   <tr>
     <th>Initial:</th>
@@ -123,7 +123,7 @@ of the position of the element. </p>
   </tr>
   <tr>
     <th>Value:</th>
-    <td><a>&lt;length&gt;</a> | <a>&lt;percentage&gt;</a></td>
+    <td><a>&lt;length-percentage&gt;</a></td>
   </tr>
   <tr>
     <th>Initial:</th>
@@ -170,7 +170,7 @@ property</h2>
   </tr>
   <tr>
     <th>Value:</th>
-    <td><a>&lt;length&gt;</a> | <a>&lt;percentage&gt;</a> | auto</td>
+    <td><a>&lt;length-percentage&gt;</a> | auto</td>
   </tr>
   <tr>
     <th>Initial:</th>
@@ -228,7 +228,7 @@ property</h2>
   </tr>
   <tr>
     <th>Value:</th>
-    <td><a>&lt;length&gt;</a> | <a>&lt;percentage&gt;</a> | auto</td>
+    <td><a>&lt;length-percentage&gt;</a> | auto</td>
   </tr>
   <tr>
     <th>Initial:</th>
@@ -285,7 +285,7 @@ property</h2>
   </tr>
   <tr>
     <th>Value:</th>
-    <td><a>&lt;length&gt;</a> | <a>&lt;percentage&gt;</a></td>
+    <td><a>&lt;length-percentage&gt;</a></td>
   </tr>
   <tr>
     <th>Initial:</th>
@@ -331,7 +331,7 @@ property</h2>
   </tr>
   <tr>
     <th>Value:</th>
-    <td><a>&lt;length&gt;</a> | <a>&lt;percentage&gt;</a></td>
+    <td><a>&lt;length-percentage&gt;</a></td>
   </tr>
   <tr>
     <th>Initial:</th>
--- a/master/painting.html
+++ b/master/painting.html
@@ -762,7 +762,7 @@ property</h3>
   </tr>
   <tr>
     <th>Value:</th>
-    <td><a>&lt;percentage&gt;</a> | <a>&lt;length&gt;</a></td>
+    <td><a>&lt;length-percentage&gt;</a></td>
   </tr>
   <tr>
     <th>Initial:</th>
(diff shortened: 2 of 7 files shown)


master/painting.html
```diff
@@ -2361,7 +2361,7 @@ <h3 id="MarkerElement">The <span class="element-name">'marker'</span> element</h
   <tr>
     <td><dfn id="MarkerWidthAttribute" data-dfn-type="element-attr" data-dfn-for="marker" data-dfn-export="">markerWidth</dfn>,
     <dfn id="MarkerHeightAttribute" data-dfn-type="element-attr" data-dfn-for="marker" data-dfn-export="">markerHeight</dfn></td>
-    <td><a>&lt;length&gt;</a> | <a>&lt;percentage&gt;</a> | <a>&lt;number&gt;</a></td>
+    <td><a>&lt;length-percentage&gt;</a> | <a>&lt;number&gt;</a></td>
```
dstorey: are these ones that use length, number, and percentage correct to use length-percentage (rather than say number-percentage)? As far as I understand it you use length-percentage if percentage computes to length.

master/propidx.html
```diff
@@ -383,11 +381,7 @@ <h1>Property Index</h1>
         </tr>
         <tr>
           <th><a>'line-height'</a></th>
-          <td>normal |
-            <a href="https://www.w3.org/TR/2008/REC-CSS2-20080411/syndata.html#value-def-number">
-              <span class="value-inst-number noxref">&lt;number&gt;</span></a> |
-	    <a>&lt;length&gt;</a> |
-	    <a>&lt;percentage&gt;</a></td>
+          <td>normal | <a>&lt;number&gt;</a> | <a>&lt;length-percentage&gt;</a></td>
```
dstorey: Note: this is different from CSS definition of line-height: normal | <number> | <length> | <percentage> | inherit, but that is probably because CSS21 is still the latest definition.

master/text.html
```diff
@@ -1987,7 +1987,7 @@ <h3 id='InlineSize'>The <span class="property">'inline-size'</span> property</h3
     </tr>
     <tr>
       <th>Value:</th>
-      <td><a>&lt;length&gt;</a> | <a>&lt;percentage&gt;</a> | <a>&lt;number&gt;</a></td>
+      <td><a>&lt;length-percentage&gt;</a> | <a>&lt;number&gt;</a></td>
```
dstorey: inline-size is also specified in Logical Properties as logical width. I think this is the same thing. in that spec it is defined as  `<width>` , which links to CSS22: `<length>` | `<percentage>` | auto | inherit  (sans number, and with auto keyword). Presumably the extra number is for compatibility with presentational attributes?

master/text.html
```diff
@@ -2472,7 +2472,7 @@ <h3 id='TextShapeMargin'>The <span class="property">'shape-margin'</span> proper
     </tr>
     <tr>
       <th>Value:</th>
-      <td><a>&lt;length&gt;</a> | <a>&lt;percentage&gt;</a></td>
+      <td><a>&lt;length-percentage&gt;</a></td>
```
dstorey: This is still `<length> | <percentage>` in CSS Shapes, but it probably just hasn't been updated yet, or should it remain as it is?

master/geometry.html
```diff
@@ -32,7 +32,7 @@ <h2 id='CX'>Horizontal center coordinate: The
   </tr>
   <tr>
     <th>Value:</th>
-    <td><a>&lt;length&gt;</a> | <a>&lt;percentage&gt;</a></td>
+    <td><a>&lt;length-percentage&gt;</a></td>
```
dstorey: These behave different to properties like width. At least in Safari/Webkit (only browsers to support them), a percentage value doesn't compute to a px value and calc(50% + 10px) for example does work but the computed value stays as  calc(50% + 10px). I'm not sure if that is a bug or not. 

Does this meet the criteria: "Equivalent to `[ <length> | <percentage> ],` where the `<percentage> `will resolve to a `<length>`." Example: "the width property can accept a `<length>` or a `<percentage>`, both representing a measure of distance. This means that `width: calc(500px + 50%);` is allowed—both values are converted to absolute lengths and added."

master/text.html
AmeliaBR: No, allowing number as a replacement for length happens automatically in all presentation attributes, see https://svgwg.org/svg2-draft/types.html#syntax. Explicitly allowing number in the CSS grammar allows it in the style property, too (this is needed for many SVG-specific properties from SVG 1). Since this is a new property, there is no legacy reason to allow plain number in the CSS property, it is probably best to harmonize with the definition in Logical Properties.

master/propidx.html
AmeliaBR: Yes, I don't think they are updating the syntax in CSS 2, and I can't find a level 3 definition of the property. But `calc(length + percentage)` works for `line-height`, so `<length-percentage>` is the correct unit type.

master/painting.html
AmeliaBR: Yes, in most places the length | number | percentage SVG properties are all lengths, with the numbers computing to lengths as "SVG user units". Allowing numbers without units in the CSS syntax is a legacy feature, we don't need to allow them in `calc()` combinations.

master/text.html
AmeliaBR: Probably should be updated there, too.

master/geometry.html
AmeliaBR: The browser implementations are bad, then. They likely have not updated their list of "applies to" elements (that is, elements for which `getComputedStyle` is supposed to return the used value instead of the inherited "computed" value).

Percentages in these properties describe lengths, they should compute to lengths, and they should be combinable with lengths in `calc`.

master/shapes.html
```diff
@@ -351,7 +351,7 @@ <h2 id="LineElement">The <span class="element-name">'line'</span> element</h2>
       </tr>
       <tr>
         <td><dfn id="LineElementX1Attribute" data-dfn-type="element-attr" data-dfn-for="line" data-dfn-export="">x1</dfn>, <dfn id="LineElementY1Attribute" data-dfn-type="element-attr" data-dfn-for="line" data-dfn-export="">y1</dfn></td>
-        <td><a>&lt;length&gt;</a> | <a>&lt;percentage&gt;</a> | <a>&lt;number&gt;</a></td>
+        <td><a>&lt;length-percentage&gt;</a> | <a>&lt;number&gt;</a></td>
```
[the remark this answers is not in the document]
AmeliaBR: Yep, part of  using the CSS syntax is wanting to allow `calc`. And that means clarifying when lengths and percentages can be combined.

AmeliaBR:
In answering those comments, I realized that the [algorithm for converting a style property syntax into a presentation attribute syntax,](https://svgwg.org/svg2-draft/types.html#syntax), will also need updating.

dstorey:
Do you meant to add an extra step between 3 and 4 for length-percentage to become length-percentage | number ?

AmeliaBR:
> Do you meant to add an extra step between 3 and 4 for length-percentage to become length-percentage | number ?

Yes, sorry, should have been more explicit.

dstorey:
@AmeliaBR if the previous comment was the only issue ,this should be ready now.

AmeliaBR:
@dstorey Can you create a separate issue with respect to https://github.com/w3c/svgwg/pull/434#discussion_r186123583 ?

PR #439, opened 2018-05-06
Fix broken and missing references

dstorey:
Also fixes a build break due to a missing < (yay XML!)


dstorey:
i'm just going to merge this as this fixes the build issue, and it is only updating links to where the fragment identifiers now live (usually related to specs moving locations)
Only real change is removing the bearing command from the at risk section, but the definition has already been removed from the spec

PR #441, opened 2018-05-08
Add myself to editors

dstorey:


dstorey:
...assuming that is ok. Feel free to reject.

dirkschulze:
Maybe we should revert all changes @dstorey did so far!!! ;)

PR #445, opened 2018-05-11
Allow paint-order to be a presentational attribute

### LINKED ISSUE #316: Add a presentation attribute for `paint-order`
In [SVG 2](https://www.w3.org/TR/SVG2/styling.html#TermPresentationAttribute) we don't list `paint-order` as a possible presentation attribute.  This was probably related to the decision not to add new presentation attributes for the ever-growing number of CSS properties that can apply to SVG.

However, every browser that supports `paint-order` supports it as a presentation attribute ([test case](https://codepen.io/AmeliaBR/pen/2033a599041bab1c59cdc51c1485017c/?editors=1100)).  So, we might as well make the spec match.

dstorey:
Fixes #316


dstorey:
I didn't change the note that new properties shouldn't be presentational attributes. It feels like this is a one off due to browser support.

Issue #82, opened 2016-04-12
Property index (appendix) needs updating


nikosandronikos:
Three issues are listed on the property index.
https://svgwg.org/svg2-draft/propidx.html

They should all be fixed in one go.
1. The table should not list properties defined in other specifications.
2. The table lacks a column for the 'computed value'.
3. The table should ideally be generated automatically from the definition of each property, to ensure that this is the complete list.

patrickdark:
Worth noting that geometry properties currently aren’t listed in the property index at all.

dirkschulze:
Beside the missing geometry properties in the prop list, there is the issue that many CSS modules do not list the SVG elements they apply to. The specs in question are at least CSS Line Layout, CSS Writing Modes and CSS 2.1. Not sure if one of the next CSS 2 erratas should list them for `display` and others. CC @fantasai @tabatkins

fantasai:
'display' has its own CSS module now. (Most properties in CSS2 are listed somewhere in CSS3 space at this point, although some of them are not actively maintained atm.) For the newer CSS modules, we should probably update the Applies to lines. I'd leave CSS2 itself untouched.

dirkschulze:
@fantasai Thanks. I'll open issue for the individual specs.

fantasai:
@dirkschulze If you want to just open a single pull request against all affected specs, that's probably sufficient. We kinda did that when we updated the format of all the Computed Value lines.

dirkschulze:
@fantasai I need to go though the entire list and will create edits (as PR) to the specs while deleting the properties from the index.

dirkschulze:
Opened issues against responsible specs. Many properties already defined "apply to: all element". Nothing to do there:
* https://github.com/w3c/csswg-drafts/issues/3411
* https://github.com/w3c/csswg-drafts/issues/3412
* https://github.com/w3c/csswg-drafts/issues/3413
* https://github.com/w3c/csswg-drafts/issues/3414

PR #446, opened 2018-05-11
Update property index

dstorey:
First part of #82.

I've added the computed value column and removed all properties that link to other specs.

I've an open question about a number of the other ones. They're often defined elsewhere, so feel like they should be removed, but link to parts of the SVG spec. They sometimes just give some historical information about how it has changed since 1.1 that doesn't seem much value can can probably be removed and the information added to the changelog (if not there already). Others Give additional rules about SVG. 1 or 2 give the definition block in SVG and link to a CSS spec (which don't always match). White-space is an example of the latter: https://svgwg.org/svg2-draft/text.html#WhiteSpaceProperty vs https://www.w3.org/TR/css-text-3/#white-space-property I think the computed value saying to see individual properties is a typo in SVG.

Some are short hands in CSS but only include the old non-shorthand values in the SVG propindex. The description in SVG usually gives the historical differences and mentions to use the new definition, making the prop index definition out of date.  I'm not sure if these should just be removed from the index and perhaps the prose.

There are some that seem exactly the same in SVG and CSS but they just apply to different elements (CSS usually says all elements while SVG restricts them). This also applies to properties that link directly to the CSS spec that I've removed, so need to work out what to do here. Perhaps the CSS specs should be updated to include SVG info for these. Or we should remove from the prop index but keep the definitions, restricting to just the differences between how they apply to HTML and SVG? (leaving historical info to the changelog as mentioned previously?)

Also some like stop-colour and stop-opacity don't have a computed value defined in the SVG spec. That seems like an oversight.

### DIFF
--- a/master/propidx.html
+++ b/master/propidx.html
@@ -61,39 +61,6 @@
           <td>yes</td>
         </tr>
         <tr>
-          <th><a>'clip'</a></th>
-          <td><a href="https://www.w3.org/TR/2008/REC-CSS2-20080411/visufx.html#value-def-shape">          <span          class="value-inst-shape noxref">&lt;shape&gt;</span></a>
-          | auto </td>
-          <td>auto</td>
-          <td><a href="coords.html#ElementsThatEstablishViewports">elements
-          which establish a new SVG viewport</a>, <a>'pattern'</a> elements and <a>'marker element'</a> elements</td>
-          <td>no</td>
-          <td>N/A</td>
-          <td><a href="https://www.w3.org/TR/2008/REC-CSS2-20080411/media.html#visual-media-group">visual</a></td>
-          <td>yes</td>
-        </tr>
-        <tr>
-          <th><a>'clip-path'</a></th>
-          <td>&lt;basic-shape&gt; | <a>&lt;url&gt;</a> |
-          none </td>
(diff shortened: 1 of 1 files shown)


master/propidx.html
```diff
@@ -388,6 +213,7 @@ <h1>Property Index</h1>
           <td>refer to font size of element itself</td>
           <td><a href="https://www.w3.org/TR/2008/REC-CSS2-20080411/media.html#visual-media-group">visual</a></td>
           <td>yes</td>
+          <td>for <a>&lt;length-percentage></a> the absolute value; otherwise as specified</td>
         </tr>
         <tr>
           <th><a>'marker property'</a></th>
```
ewilligers: I think this should simply be 'marker'.

master/propidx.html
```diff
@@ -498,6 +322,7 @@ <h1>Property Index</h1>
           <td>N/A</td>
           <td><a href="https://www.w3.org/TR/2008/REC-CSS2-20080411/media.html#visual-media-group">visual</a></td>
           <td>yes</td>
+          <td></td>
```
ewilligers: In a future PR, we should specify the computed value for stop-opacity (by copying off stroke-opacity).

PR #452, opened 2018-05-14
Remove mesh gradients

dstorey:


dstorey:
These will probably all need an additional sweep to remove all the demo files and illustrations from the source that are no longer linked. We can probably do that all at once as a clean up task.

AmeliaBR:
For the changelog, for this and the hatch PR, maybe make it explicit that this is a "defer" instead of a "remove for ever"?

dstorey:
How about now?

dstorey:
New bug is #454  and I've created the SVG2/21 tags

dstorey:
Merging these two as I just realized Dirk had approved them. We can massage the wording if they need changing all together in a new PR.

PR #453, opened 2018-05-14
Remove SVGCursorElement

dstorey:


dstorey:
Something weird has happened with this PR as it is counting the whole file as changed. I'll recreate it when I have some time later today or tomorrow.

PR #455, opened 2018-05-15
Remove SVGCursorElement...

dstorey:
...and the ability to specify a cursor element from the CSS cursor property.


dstorey:
I'm not sure if I was too aggressive with the removals in this PR, but it looks like the reason for the Cursor section was for the cursor element and to extend CSS cursor to be able to specify a cursor element. Without the cursor element it feels like that entire section is not needed, as cursor CSS property just behaves like in HTML and the CSS spec, so I chose to remove it entirely.

master/definitions.xml
```diff
@@ -581,7 +568,7 @@
   <attributecategory
     name='presentation'
     href='styling.html#TermPresentationAttribute'
-    presentationattributes='alignment-baseline, baseline-shift, clip, clip-path, clip-rule, color, color-interpolation, color-interpolation-filters, color-rendering, cursor, direction, display, dominant-baseline, fill, fill-opacity, fill-rule, filter, flood-color, flood-opacity, font-family, font-size, font-size-adjust, font-stretch, font-style, font-variant, font-weight, glyph-orientation-horizontal, glyph-orientation-vertical, image-rendering, letter-spacing, lighting-color, marker-end, marker-mid, marker-start, mask, opacity, overflow, paint-order, pointer-events, shape-rendering, solid-color, solid-opacity, stop-color, stop-opacity, stroke, stroke-dasharray, stroke-dashoffset, stroke-linecap, stroke-linejoin, stroke-miterlimit, stroke-opacity, stroke-width, text-anchor, text-decoration, text-rendering, transform, unicode-bidi, vector-effect, visibility, word-spacing, writing-mode'/>
```
dstorey: I guess I should have left this in? Is the cursor attribute the same as the CSS cursor property?

master/propidx.html
```diff
@@ -124,19 +124,6 @@ <h1>Property Index</h1>
           <td><a href="https://www.w3.org/TR/2008/REC-CSS2-20080411/media.html#visual-media-group">visual</a></td>
           <td>yes</td>
         </tr>
-        <tr>
```
dstorey: I figured this isn't needed as it is defined as in the CSS spec now. It does limit what elements cursor can be used on, while CSS states all elements, but that is a common problem of most CSS properties not defined in SVG.

master/definitions.xml
AmeliaBR: Yes, we still need to define somewhere that you can specify the cursor style property as a presentation attribute.

master/definitions.xml
```diff
@@ -700,7 +687,6 @@
   <property name='color' href='painting.html#ColorProperty'/>
   <property name='color-interpolation' href='painting.html#ColorInterpolationProperty'/>
   <property name='color-rendering' href='painting.html#ColorRenderingProperty'/>
-  <property name='cursor' href='interact.html#CursorProperty'/>
```
AmeliaBR: There still needs to be a `cursor` property definition for auto-linking, but it should have an updated `href` pointing to CSS-UI.

master/interact.html
```diff
@@ -38,10 +38,6 @@ <h2 id="Introduction">Introduction</h2>
   attribute on the <a>'svg'</a> element and on the
   characteristics of the user agent, users are able to zoom
   into and pan around SVG content.</li>
-
```
AmeliaBR: No reason to remove this. The `cursor` property still exists, and is still part of interactive SVG.

master/definitions.xml
dstorey: added it back

master/propidx.html
AmeliaBR: That table includes other properties that are defined elsewhere but apply to SVG elements and have SVG presentation attributes (e.g., `display`, `font-family`), so I'd leave it there.

master/definitions.xml
dstorey: Done

master/propidx.html
dstorey: I thought we planned to remove all the properties not defined in SVG in the property index, but maybe I misunderstood and all the ones that have a presentational attribute should stay?

master/linking.html
```diff
@@ -679,7 +678,6 @@ <h4 id="processingURL-validity">Valid URL targets</h4>
   <!--
   <li>the <a href="color.html#ColorProfileSrcProperty">'src'</a> descriptor on an @color-profile definition must reference an ICC profile resource</li>
   -->
-  <li>the <a>'cursor property'</a> property must reference either a <a>'cursor element'</a> element or a document that can be processed as an image</li>
   <li>the <a>'fill'</a> property (see <a href="painting.html#SpecifyingPaint">Specifying paint</a> for reference rules)</li>
   <li>the <a>'marker property'</a>, <a>'marker-start'</a>, <a>'marker-mid'</a> and <a>'marker-end'</a> properties must reference a <a>'marker element'</a> element.</li>
   <li>the <a>'shape-inside'</a> and <a>'shape-subtract'</a> properties must reference an element type
```
AmeliaBR: There's another reference to the cursor property accepting URL references to cursor elements in the paragraph after this list that needs to be removed.

master/interact.html
dstorey: How much of this do you want to keep in?

AmeliaBR:
Overall: I agree that we don't need a full section about the `cursor` property, since we're removing all SVG-specific behavior. 

However, the `cursor` property still needs to be a presentation attribute, and needs to be in the [required properties list](https://svgwg.org/svg2-draft/styling.html#RequiredProperties) (no change needed, there; CSS UI is already referenced so just need to update the auto-link from definitions).  See comments on specific edits for those.

Otherwise, looks good, but probably needs a more careful review & maybe a test build to find broken cross-links.

master/linking.html
dstorey: done

master/interact.html
AmeliaBR: Meh, it's pretty light-weight content anyway. `cursor` is no more essential for interactivity than `:hover` styles are. Might as well remove it. Ignore my previous comment.

master/propidx.html
AmeliaBR: I think keeping all the presentation attribute properties in an appendix is reasonable.

But regardless, if we agreed to remove them, it should be in a separate PR cleaning up the entire table, not here.

master/interact.html
dstorey: Does that mean it is approved to be merged now?

PR #464, opened 2018-05-18
Remove defintion of white-space

### LINKED ISSUE #447: Computed value for white-space incorrect?
As spotted in #446

Says the property is defined in CSS but then defines in in SVG as well. Includes a computed value of "see individual properties" but it isn't a short hand. CSS defines it as "as specified"

As this section is mostly a history lesson around the property replacing xml:space, should it just all be removed completely? The only thing special I see is that it only applies to text elements rather than all elements (that seems fairly common for other things defined in CSS like font etc that don't have their own section). The example also lightly makes fun of Firefox.

Perhaps just keep the xml:space section and mention in a note that one should use white-space instead of this deprecated attribute?

dstorey:
Fixes #447

The spec redefines white-space, which is defined in CSS Text Level 3 (which is also pointed out in the note). Removing this means the incorrect computed value is removed and makes sure implementers/devs use the CSS version.

Also removed the compat info about Firefox.

### DIFF
--- a/master/definitions.xml
+++ b/master/definitions.xml
@@ -748,7 +748,6 @@
   <property name='shape-subtract'  href='text.html#TextShapeSubtract'/>
   <property name='shape-padding'   href='text.html#TextShapePadding'/>
   <property name='text-overflow'   href='text.html#TextOverflowProperty'/>
-  <property name='white-space'     href='text.html#WhiteSpaceProperty'/>
 
   <!-- ... font properties defined in CSS Fonts 3 -->
   <property name='font-kerning'     href='https://www.w3.org/TR/css-fonts-3/#font-kerning-prop'/>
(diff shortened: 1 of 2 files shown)


dstorey:
I'm not 100% sure if the definition link in definitions.xml should move to CSS Text where I moved it or stay where it was but link to CSS Text instead of an internal link to the definition I removed. 

It is described in the spec but the description in limited to what it is (presumably doesn't add anything not in CSS Text) and a SVG specific example.

```
  <h4 id="TextWhiteSpace">SVG 2 Preferred white space handling, the <span class='property'>'white-space'</span> property</h4>

  <p>
    Rendering of white space in SVG 2 is controlled by the <a>'white-space'</a>
    property. This specifies two things:
  </p>
  <ul>
    <li>whether and how white space inside the element is collapsed</li>
    <li>whether lines may wrap at unforced soft wrap opportunities</li>
  </ul>
```

PR #465, opened 2018-05-23
Export text content element definition

dirkschulze:


master/text.html
```diff
@@ -217,7 +217,7 @@ <h3 id="Definitions">Definitions</h3>
       produce legible text in small sizes.
     </dd>
 
-    <dt id="TermTextContentElement">text content element</dt>
+    <dt><dfn id="TermTextContentElement" data-dfn-type="dfn" data-dfn-export="">text content element</dfn></dt>
```
dstorey: This should be data-export=""

dirkschulze: right

PR #471, opened 2018-06-18
Remove unneeded 'shape-outside' definition. Fixes #288.

### LINKED ISSUE #288: Text chapter and definitions.xml still reference shape-outside
The note in [11.4. Text layout – Content Area](https://www.w3.org/TR/2016/CR-SVG2-20160915/text.html#TextLayoutContentArea) has an example that still uses the term 'shape-outside'. This should be changed to [shape-subtract](https://www.w3.org/TR/2016/CR-SVG2-20160915/text.html#TextShapeSubtract).

'shape-outside' also remains in the definitions.xml file.

Tavmjong:
No visible change in spec.


ewilligers:

PR #476, opened 2018-06-28
Textpath pathlength

### LINKED ISSUE #382: Excess parenthesis in text layout algorithm? ("Position on path" part)
In the "11.5. Text layout – Algorithm" [1], step 8.5.1.2.4, which reads:

> Let offset be the value of the ‘textPath’ element's ‘startOffset’ attribute, adjusted due to any ‘pathLength’ attribute on the referenced element (if the referenced element is a ‘path’ element).

there's an additional restriction placed on when 'pathLength' applies ("if the referenced element is a ‘path’ element"). This restrictions seems to stem from the initial landing of the text layout algorithm. However (IIRC), since then all the basic shape elements gained 'pathLength' powers. Maybe this restriction (parenthesis) should be dropped?

[1] https://svgwg.org/svg2-draft/text.html#TextLayoutAlgorithm

Tavmjong:
Remove unintended restriction that 'pathLength' only applied to paths and not shapes in regard to 'startOffset'.

Fixes #382.

### DIFF
--- a/master/changes.html
+++ b/master/changes.html
@@ -546,9 +546,15 @@ have been made.</p>
 
 <div class='changed-since-cr1'>
 <ul>
+
   <li>Specify <a>DOMPointInit</a> as argument type for getCharNumAtPosition() instead of
     <a>DOMPoint</a>.
   </li>
+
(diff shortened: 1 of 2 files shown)


ewilligers:
The changes to `master/changes.html` and `master/text.html` are fine. The other clean ups should be split out.

PR #477, opened 2018-06-29
remove redundant words

prlbr:
I’ve checked for a few common words whether they occur twice in a row, which is a common typo, and fixed these.


prlbr:

PR #481, opened 2018-07-05
Clarify that mesh and hatch have been deferred

ewilligers:
There were references in a couple of chapters' changes to
mesh and hatch. Each such chapter now notes that mesh and
hatch have been deferred.

### DIFF
--- a/master/changes.html
+++ b/master/changes.html
@@ -317,6 +317,7 @@ have been made.</p>
 <div class='changed-since-cr1'>
 <ul>
   <li>Corrected incorrect interface name for definition of <a>SVGUseElementShadowRoot</a>. <a href="https://github.com/w3c/svgwg/issues/290">Github #290</a>.</li>
+  <li>Deferred mesh element to a future version of SVG.</li>
 </ul>
 </div>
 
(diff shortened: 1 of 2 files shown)


master/changes.html
```diff
@@ -317,6 +317,7 @@ <h3 id="structure">Document Structure chapter (owner: BogdanBrinza)</h3>
 <div class='changed-since-cr1'>
 <ul>
   <li>Corrected incorrect interface name for definition of <a>SVGUseElementShadowRoot</a>. <a href="https://github.com/w3c/svgwg/issues/290">Github #290</a>.</li>
+  <li>Deferred mesh element to a future version of SVG.</li>
```
dirkschulze: That is the changes section that explains the changes from the previous REC to the next version IIRC. Not from a previous publication to the new one. So we probably should not mention it.

AmeliaBR: The changes section covers changes between all TR publications, with changes from the most recent TR highlighted (and the list includes lots of other changes that were made and then changed again).

AmeliaBR:
I personally don't feel that the clarifications are necessary. We already have the main change noted in the relevant chapter. What I did for another situation when tidying the appendix was to modify the original change entry to mention that it was later reverted:

https://github.com/w3c/svgwg/blob/a4a5a206daee1b1f28140582ace93d8e069b9e2a/master/changes.html#L684

ewilligers:
Closing as reviewers have explained preference for the existing text.

PR #482, opened 2018-07-05
Inline size: remove 'number' value.

### LINKED ISSUE #436: Harmonise inline-size with CSS Logical properties
from @AmeliaBR 

>  allowing number as a replacement for length happens automatically in all presentation attributes, see https://svgwg.org/svg2-draft/types.html#syntax. Explicitly allowing number in the CSS grammar allows it in the style property, too (this is needed for many SVG-specific properties from SVG 1). Since this [inline-size] is a new property, there is no legacy reason to allow plain number in the CSS property, it is probably best to harmonize with the definition in Logical Properties.

Tavmjong:
Fixes #436

### DIFF
--- a/master/changes.html
+++ b/master/changes.html
@@ -549,6 +549,7 @@ have been made.</p>
   <li>Specify <a>DOMPointInit</a> as argument type for getCharNumAtPosition() instead of
     <a>DOMPoint</a>.
   </li>
+  <li>Remove 'number' value from <a>'inline-size'</a> property for constency with CSS.</li>
 </ul>
 </div>
 
--- a/master/text.html
+++ b/master/text.html
@@ -1987,7 +1987,7 @@
     </tr>
     <tr>
       <th>Value:</th>
-      <td><a>&lt;length-percentage&gt;</a> | <a>&lt;number&gt;</a></td>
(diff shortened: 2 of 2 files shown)


ewilligers:
Why isn't auto supported?

Tavmjong:
What would be the 'auto' value?

ewilligers:
We seem to have SVG and CSS defining `inline-size` in conflicting ways.

https://drafts.csswg.org/css-logical/#propdef-inline-size
supports 'auto' and has 'auto' as the initial value.

> What would be the 'auto' value?

Discussed for HTML in https://www.w3.org/TR/CSS21/visudet.html#propdef-width

Perhaps the SVG spec needs to discuss it too.


I don't think we have any browser implementations following the SVG grammar. Blink supports 'auto' and has 'auto' as the initial value.

AmeliaBR:
@ewilligers Can you make a separate issue for this, since this PR has now been merged? `auto` should be allowed, and should be the default value, meaning do not constrain the inline size at all.  Although that brings up the question of what the computed value should be: the text length, I guess.

There should maybe also be a note that says that the definition in CSS Logical Properties will supersede the SVG-specific one, when that spec is finalized.

PR #493, opened 2018-07-10
Mark SVGNameList as an example

saschanaz:
This change:

1. makes it visually look like an example
2. enables IDL crawlers to exclude it.


saschanaz:

PR #495, opened 2018-07-10
Use class="extract" to exclude IDL

saschanaz:
This is what bikeshed does e.g. in [CSSOM](https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-_webkit_cased_attribute). It doesn't remove the flag so that IDL crawlers can also detect it.

### DIFF
--- a/master/struct.html
+++ b/master/struct.html
@@ -2771,7 +2771,7 @@ Authors are encouraged to use the <a href="https://www.w3.org/TR/dom/#dom-docume
 Other SVG implementations must support the following IDL fragment.
 </p>
 
-<pre class="idl" edit:excludefromidl="true"><span class="comment">// must only be implemented in certain implementations</span>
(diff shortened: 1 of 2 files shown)


saschanaz:

AmeliaBR:
Whoops! This is breaking the build.

(The _processing.js_ scripts don't run on a full DOM objects, they run on a pseudo-DOM created with [the `xmldom` node module](https://www.npmjs.com/package/xmldom), and don't have a `classList` object. Easy mistake to make @saschanaz, and thanks for the contribution nonetheless.)

Anyway: I'll fix it, but in future, @dirkschulze, please don't merge any changes to the _tools/_ files without running a test build!

saschanaz:
Whoops, sorry!

BTW xmldom is being poorly maintained, perhaps we can use jsdom fragment mode instead to get proper DOM APIs.

AmeliaBR:
See https://github.com/w3c/svgwg/pull/539 for the fixes.

PR #521, opened 2018-08-10
Clean-up Changes appendix

AmeliaBR:
Tidying up the changes appendix. See commit messages for details.

Shouldn't really be any normative changes here, but I'd appreciate someone else taking a quick look to make sure I haven't broken anything.

I suspect that I'll find a few more missing changes as I go through the rest of the implementation review. But, it's still probably better to merge this sooner rather than later, so that other PRs (which all change the changes appendix, of course) can be rebased over top.

Other PRs will also need minor edits to add links to issues and PRs, as well as annotations to indicate that they are changes since the August CR, so that the tidy-up can last more than a few days!


Tavmjong:
Good work! I also agree with the decision on z-index.

PR #527, opened 2018-08-14
Make examples xml valid.

JohnLukeBentley:
Make the "pre-formatted text" and "Text on a path" xml valid by shifting the ">" of the svg element after the attributes.

This follows a recent similar correction I made to another nearby example.


ewilligers:

PR #528, opened 2018-08-16
Language tags are case-insensitive. Closes #517

### LINKED ISSUE #517: systemLanguage is not clear about case-insensitive matches
Document structure chapter
Switch and systemLanguage
https://svgwg.org/svg2-draft/struct.html#SwitchElement

The systemLanguage attribute is a comma-separated list of BCP 47 langtags.

The SVG 2.0 spec states the systemLanguage attribute:

> Evaluates to "true" if one of the languages indicated by user preferences exactly equals one of the languages given in the value of this parameter, or if one of the languages indicated by user preferences exactly equals a prefix of one of the languages given in the value of this parameter such that the first tag character following the prefix is "-".

The phrase "exactly equals" has misled many implementers. It would be better to state explicitly that the comparison is a **case-insensitive match**. For example, the langtag "en-US" is equivalent to "en-us". To most readers, it is not "exactly equals" but is "equivalent". Saying it is a case-insensitive match would just make it clear.

Clarity is an issue. Chrome and Firefox browsers currently use **case-sensitive** matches for systemLanguage. Consequently, systemLanguage="EN" evaluates to false but systemLanguage="en" evaluates to true on those browsers.

- https://jsfiddle.net/35yL6qnw/5/

BCP 47 (https://tools.ietf.org/html/bcp47) implies that the case of the langtag does not matter when doing some operations, but doesn't actually make a clear statement that case does not matter for any comparison. You can search through the RFCs and find mention that case does not matter in some situations, but I did not find the broad statement.

Search the BCP 47 document for "insensitive" to find statements such as
> At all times, language tags and their subtags, including private use and extensions, are to be treated as case insensitive: there exist conventions for the capitalization of some of the subtags, but these MUST NOT be taken to carry meaning.
and
> Language tags and thus language ranges are to be treated as case-insensitive: there exist conventions for the capitalization of some of the subtags, but these MUST NOT be taken to carry meaning. Matching of language tags to language ranges MUST be done in a case-insensitive manner.

A better statement for the SVG specification would be (also using "language tag" instead of "language"):
>  Evaluates to "true" if one of the **language tags** indicated by user preferences is a **case-insensitive match** to one of the **language tags** given in the value of this parameter, or if one of the **language tags** indicated by user preferences is a **case-insensitive prefix** of one of the languages given in the value of this parameter such that the first tag character following the prefix is "-".

Interestingly, BCP 47 (RFC 4647 section 3) states
> Protocols and specifications requiring conformance to this specification MUST clearly indicate the particular mechanism used in selecting or matching language tags.
SVG is a specification, but it has not clearly indicated the mechanism used in matching langtags. SVG 1.1 uses 3.3.1 Basic Filtering; there's a match if the basic filtered set is not empty. The SMIL allowReorder semantics are more complicated (and even ambiguous).

GLRoylance:
Changed description of systemLanguage matching to make it clear that it is a case-insensitive match rather than an "exact match".

### DIFF
--- a/master/struct.html
+++ b/master/struct.html
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional+edit//EN" "xhtml1-transitional+edit.dtd">
 <html lang="en" xmlns="http://www.w3.org/1999/xhtml" xmlns:edit="http://xmlns.grorg.org/SVGT12NG/">
 <head>
@@ -1647,11 +1647,11 @@ versions of a given extension.</p>
   </dd>
 </dl>
 
-<p>Evaluates to "true" if one of the languages indicated by
-user preferences exactly equals one of the languages given in
-the value of this parameter, or if one of the languages
-indicated by user preferences exactly equals a prefix of one of
-the languages given in the value of this parameter such that
+<p>Evaluates to "true" if one of the language tags indicated by
+user preferences is a case-insensitive match of one of the language tags given in
+the value of this parameter, or if one of the language tags
+indicated by user preferences is a case-insensitive prefix of one of
+the language tags given in the value of this parameter such that
 the first tag character following the prefix is "-".</p>
 
 <p>Evaluates to "false" otherwise.</p>


ewilligers:
My informal testing suggests Chrome and Firefox are case-sensitive, and Edge is case-insensitive.

Has this been discussed by the Working Group?

GLRoylance:
It does not matter what Chrome or Firefox thinks; it matters what BCP 47
thinks; BCP 47 says case insensitive. I speak "en-US", and you want SVG 2
to say that language is different from "en-us"? BCP 47 says the preferred
capitalization is "en-US", but comparisons do not matter.

But Chrome and Firefox recognize that the comparison should be
case-insensitive.

Chrome/Blink is working to make systemLanguage case-insensitive. See
https://bugs.chromium.org/p/chromium/issues/detail?id=869440 and the fsoder
comment on issue #517.

Firefox's nightlies are now case-insensitve. See
https://bugzilla.mozilla.org/show_bug.cgi?id=1480946 and the longsonr
comment on issue #517.

And CSS says the :lang psuedo selector does case insensitive comparisons.
W3C sez

The pseudo-class ':lang(C)' matches if the element is in language C.
Whether there is a match is based solely on the identifier C being either
equal to, or a hyphen-separated substring of, the element's language value,
in the same way as if performed by the '|='
<https://www.w3.org/TR/CSS21/selector.html%23id-selectors#attribute-selectors>
 operator. *The matching of C against the element's language value is
performed case-insensitively for characters within the ASCII range.* The
identifier C does not have to be a valid language name.
https://www.w3.org/TR/CSS21/selector.html%23id-selectors#lang

The pseudo-class :lang(C) represents an element that is in language C.
Whether an element is represented by a :lang() selector is based solely on
the element's language value (normalized to BCP 47 syntax if necessary)
being equal to the identifier C, or beginning with the identifier C
immediately followed by "-" (U+002D). *The matching of C against the
element's language value is performed case-insensitively within the ASCII
range.* The identifier C does not have to be a valid language name.
https://drafts.csswg.org/selectors-3/#lang-pseudo

(some emphasis added)

AmeliaBR:
I've added a changelog entry, and I think this is good to merge.

@svgeesus Are we ok marking this non-substantive for IP? It's a normative correction, but not really adding any intellectual property.

Alternatively, @GLRoylance, if you are a W3C contributor, please link up your W3C & Github accounts following the instructions here: https://labs.w3.org/hatchery/repo-manager/pr/id/w3c/svgwg/528

GLRoylance:
There ain't no intellectual property here. The edit only clarifies what BCP 47 says.
I release my pull request as CC0.
I am not a member of any W3C entity.

ewilligers:

Issue #529, opened 2018-08-17
stroke-width initial value must include units


sideshowbarker:
At https://svgwg.org/svg2-draft/painting.html#StrokeWidthProperty, the initial value must not be just `1`; instead the units must be explicitly included — so it needs to be `1px`.

That’s per the chain of syntax requirements from https://drafts.csswg.org/css-values-4/#typedef-length-percentage to https://drafts.csswg.org/css-values-4/#length-value to https://drafts.csswg.org/css-values-4/#dimension that the SVG spec normative includes by reference (due to citing https://www.w3.org/TR/css-values/#typedef-length-percentage) — which indicate that a `<length-percentage>` value must include units if it’s a length.

See also https://drafts.fxtf.org/fill-stroke/#stroke-width, which correctly has the initial value as `1px` instead of just `1`.

AmeliaBR:
The stroke-width syntax should actually be `<length-percentage> | <number>`, since plain numbers are valid in CSS properties, not only presentation attributes. Must have got dropped when converting to the `<length-percentage>` syntax.

But I don't see a problem with making the initial value have units, since the computed value is always converted to absolute length anyway.

ewilligers:
> The stroke-width syntax should actually be `<length-percentage> | <number>`, since plain numbers are valid in CSS properties, not only presentation attributes.

[Confirmed](https://docs.google.com/document/d/1aOJTvNCWlGpckxEbGd_zylj7Nz_kgLmrp7TXz8Wm8Fk/edit) support for plain numbers in all browsers. There are a couple of implementation bugs: Only Safari changes "10" to "10px" when serializing. Only Firefox gives "10" as the computed style. Neither has these bugs for line-height.

sideshowbarker:
> The stroke-width syntax should actually be `<length-percentage> | <number>`, since plain numbers are valid in CSS properties, not only presentation attributes.

I’m not sure what you mean by *“plain numbers are valid in CSS properties”* but as far as the CSS `stroke-width` property specifically, unit-less numbers aren’t valid values for it per https://www.w3.org/TR/fill-stroke-3/#stroke-width and https://drafts.fxtf.org/fill-stroke/#stroke-width, which both define the allowed value as `<length-percentage>`.

> Confirmed support for plain numbers in all browsers.

It’s unsurprising that browsers accept and process unit-less numbers in the value. The issue here isn’t about the processing behavior/requirements in browsers; instead it’s about the document-conformance requirements — the authoring rules.

[...]

PR #530, opened 2018-08-17
Change stroke-width initial value to 1px

sideshowbarker:
Fixes https://github.com/w3c/svgwg/issues/529


sideshowbarker:

PR #539, opened 2018-09-04
Revert & replace changes from PR #495

AmeliaBR:
Reverts commit 63a42e41f0009ef737ab9f58f83e689ed88007ef

Creates the same effect using the build system:
elements with the `edit:excludefromidl="true"` attribute in the source
are given the class "extract" in the build.
Which apparently has an effect in the Reffy system
(for parsing the IDL of different specs),
though I'm not sure exactly what.
See https://github.com/tidoust/reffy/issues/123

### DIFF
--- a/tools/publish/processing.js
+++ b/tools/publish/processing.js
@@ -419,8 +419,9 @@ function doCompleteIDL(conf, page, n) {
     utils.forEachNode(doc, function(n) {
       if (n.nodeType == n.ELEMENT_NODE &&
           n.localName == "pre" &&
-          n.getAttribute("class") == "idl") {
-        if (n.classList.contains("extract")) {
+          /\bidl\b/.test(n.getAttribute("class")) ) {
+        if (n.svg_excludefromidl) {
+          delete n.svg_excludefromidl;
           return;
         }
         if (idl.length) {
@@ -873,6 +874,12 @@ exports.formatMarkup = function(conf, page, doc) {
         }
       }
       n.removeAttribute("edit:toc");
+      if (n.hasAttribute("edit:excludefromidl")) {
+        n.svg_excludefromidl = true;
+        n.setAttribute("class", n.getAttribute("class")+" extract");
+        //`extract` class is used by Reffy when building IDL indexes
+        n.removeAttribute("edit:excludefromidl");
+      }
     }
   });
 
--- a/master/struct.html
+++ b/master/struct.html
@@ -2767,7 +2767,7 @@ Authors are encouraged to use the <a href="https://www.w3.org/TR/dom/#dom-docume
 Other SVG implementations must support the following IDL fragment.
 </p>
 
-<pre class="idl extract"><span class="comment">// must only be implemented in certain implementations</span>
+<pre class="idl" edit:excludefromidl="true"><span class="comment">// must only be implemented in certain implementations</span>
 partial interface <a>Document</a> {
   readonly attribute DOMString title;
   readonly attribute DOMString referrer;


AmeliaBR:
@saschanaz @foolip 

I couldn't figure out what effect the `extract` class actually has. Nothing in the Reffy code (https://github.com/tidoust/reffy/) seems to look for a class of that name.

That said, there's no harm in adding it, and I'd really like to merge this PR (and fix the SVG build system) sooner rather than later, so if you don't get back to me in the next ~24hours with a suggestion to approach this in a different way, I'll be merging this PR. We can also make further changes later.

saschanaz:
>I couldn't figure out what effect the extract class actually has.

`extract` is [to align with Bikeshed](https://tabatkins.github.io/bikeshed/#no-idl) so that we can use same detection algorithm to exclude IDLs for each spec.

BTW, can we just replace `classList.contains` with `className.includes`, rather than reintroducing the `edit:` thing?

AmeliaBR:
> BTW, can we just replace classList.contains with className.includes, rather than reintroducing the edit: thing?

We probably could, but I decided to err on the side of "don't touch the legacy code anymore than we have to".

saschanaz:
That makes sense, but the previous change was short enough... Anyway I'm okay with this change if we keep the `.extract` thing.

foolip:
> BTW, can we just replace classList.contains with className.includes, rather than reintroducing the edit: thing?

`contains` is the right thing here, see https://dom.spec.whatwg.org/#domtokenlist.

I think I've spotted the problem, will send a PR.

saschanaz:
>`contains` is the right thing here,

It's right, but somehow xmldom does not support it.

foolip:
> It's right, but somehow xmldom does not support it.

Ah, I see.

tools/publish/processing.js
```diff
@@ -419,8 +419,9 @@ function doCompleteIDL(conf, page, n) {
     utils.forEachNode(doc, function(n) {
       if (n.nodeType == n.ELEMENT_NODE &&
           n.localName == "pre" &&
-          n.getAttribute("class") == "idl") {
```
foolip: I said I spotted the problem and would send a PR. The problem in https://github.com/w3c/svgwg/pull/495 was in these two lines. `n.getAttribute("class") == "idl"` means that `n.classList.contains("extract")` is always going to be false.

However, I thought this was a plain revert and it is not. This keeps `edit:excludefromidl`, adds "extract" do that IDL block and excludes it from the IDL summary.

It's probably doable to use the extract class in the source as well, but that's an editorial decision that I'm happy to not have an opinion on.

tools/publish/processing.js
```diff
@@ -419,8 +419,9 @@ function doCompleteIDL(conf, page, n) {
...
-        if (n.classList.contains("extract")) {
+          /\bidl\b/.test(n.getAttribute("class")) ) {
```
saschanaz: We probably can use `.className` here instead.

tools/publish/processing.js
AmeliaBR: Yes, that was an extra point of failure that I noticed (and why I switched it to use a regex test instead), although the main point of failure was still trying to use `classList` and causing the entire script to abort.

tools/publish/processing.js
AmeliaBR: I don't think xmldom exposes `className` (unless they've added it and not changed their docs).

As you mentioned elsewhere, xmldom might not be the most ideal tool to use, but I'm trying to minimize changes to the tools. When we do change the SVG build system, it will to be replace it with a well-maintained alternative like Bikeshed.

tools/publish/processing.js
AmeliaBR: As for `edit:excludefromidl`, I was mostly concerned that it might be used somewhere else (these build tools are also used by other draft specs), to minimize possible side effects.
(Although I just ran a search & it seems to have been a one-off.) 

But since the end result works for everyone, I'm going to merge & we should get an updated build on svgwg.org within an hour.

PR #540, opened 2018-09-04
Reintroduce segment-completing closepath as a theoretical concept

AmeliaBR:
Because we use it to define the equivalent paths of basic shapes.

I've changed the terminology to refer to the segment-completing close path as an _operation_ instead of a _command_, and have added a short sub-section defining it.

I also included a very short note about the proposal for adding it to the path data syntax, in a future version of the specification.

This fixes build problems introduced by PR #398.

@ewilligers can you take a look at the new text I've added & let me know if you think it works to patch over the disconnect.


ewilligers:
lgtm

PR #542, opened 2018-09-04
 Add stub files for removed chapters/appendices

AmeliaBR:
As resolved in the [13 August 2018 teleconference](https://www.w3.org/2018/08/13-svg-minutes.html#resolution01):

> RESOLUTION: If possible, add stub pages to spec for SVG 1.1 chapters that no longer exist, redirecting to correct specs for the content.

So that links to URLs of the form https://www.w3.org/TR/SVG/backward.html won't 404.

Where relevant, the stubs include a very brief description of what happened to the content, with links to its current location (e.g., another chapter or a CSS spec). There are also links back to the SVG11 TR version of the chapter, and to the main SVG2 Overview page.

To include these files in the final build, without messing up building the actual spec, they are included as "resource" files in publish.xml. This means that they are copied as-is to the final directory, with no build system modifications.

_________________________________

Note: this branch includes the changes from PR #539 and #540, since I needed to get a clean test build running locally. **Those PRs should be merged first.**

### DIFF
--- a/master/backward.html
+++ b/master/backward.html
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE html>
+<html lang="en" xmlns="http://www.w3.org/1999/xhtml">
+<head>
+  <title>Removed: Backwards Compatibility — SVG 2</title>
+  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
+  <meta name="viewport"           content="width=device-width, initial-scale=1, shrink-to-fit=no"/>
+  <link rel="stylesheet" href="//www.w3.org/StyleSheets/TR/2016/base.css" type="text/css" media="screen">
+</head>
+<body>
+  <h1>Removed: Backwards Compatibility</h1>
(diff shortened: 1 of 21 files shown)


AmeliaBR:
@svgeesus
You weren't on the call we discussed this, Chris: Do you think including these stub files in the TR published build is OK from the perspective of the W3C publication process? Any suggested changes to the wording/formatting?

svgeesus:
@AmeliaBR 

> Do you think including these stub files in the TR published build is OK from the perspective of the W3C publication process? Any suggested changes to the wording/formatting?

Yes, I think this would be fine. It's a nice way to avoid cluttering up the single-page spec while also avoiding broken inbound links.

AmeliaBR:
Thanks Chris.

To fix the /TR/ broken links, we'll need another CR update. We can talk on the next call about whether there are any other quick fixes / cleanup we want to integrate before that. But it should hopefully be a fairly straightforward publication process only a month after the last one.

AmeliaBR:
Agenda+ so we don't forget to have that talk about republishing!

Issue #514, opened 2018-08-04
References needs a thorough overhaul


svgeesus:
The references all need to be checked for

- actual use in the spec
- normative vs. informative
- links to latest published version

This sort of thing is done automatically by tools like bikeshed, but needs to be periodically done manually here.

See also https://github.com/w3c/svgwg/issues/71 https://github.com/w3c/svgwg/issues/227 https://github.com/w3c/svgwg/issues/365

svgeesus:
Hopefully fixed by https://github.com/w3c/svgwg/pull/515

svgeesus:
Normative Refs includes [Document Object Model (DOM) Level 2 Views Specification](https://www.w3.org/TR/DOM-Level-2-Views/)

> Document Status Update 2011-09-22: This paragraph is informative. The concepts this document defines are obsolete. The 'document' and 'defaultView' attributes are defined in the HTML5 specification with simplified semantics. The Web Applications Working Group encourages implementation of these concepts as defined by HTML5.

I don't find DOM Views in HTML5 (or in DOM LS either).

svgeesus:
I don't understand why we have references to both HTML 5.1 2nd Edition and also HTML LS. Can refs to the former simply be converted to reference the latter? Is there some specific W3C-only functionality that we need (and is implemented)?

svgeesus:
W3C URL spec [says](https://www.w3.org/TR/url-1/)

> Work on this document has been discontinued and it should not be referenced or used as a basis for implementation.

Replacing with direct link to URL LS.

svgeesus:
Do we really still have a normative dependency on **Synchronized Multimedia Integration Language (SMIL 3.0)** ?

svgeesus:
Wondering whether WOFF 2 should be added, alongside WOFF 1?

AmeliaBR:
> I don't understand why we have references to both HTML 5.1 2nd Edition and also HTML LS.

Because no one has gone through and updated all the references & links to W3C HTML to point to their matching places in WHATWG.  I don't think there are any details we need to reference that are only in the W3C versions.

> Do we really still have a normative dependency on Synchronized Multimedia Integration Language (SMIL 3.0) ?

It looks like we still have definitions of SMIL-related events in interact.html. (Probably should see if those can be safely moved to the separate animations module.)

The other thing that is borrowed from SMIL is `<switch>`, and we have a normative reference for the `allowReorder` behavior. So at the very least the ref should stay for that.

[...]

PR #544, opened 2018-09-17
Change from HTML 5.1 to HTML LS, #514

svgeesus:
Update all internal links to [HTML] which were linking to HTML 5.1 in the refs (with the wrong bibliographic reference). Actual [HTML51] was not found.

### DIFF
--- a/master/render.html
+++ b/master/render.html
@@ -638,7 +638,7 @@ positions required on the output device. Resampling
 requirements are discussed under
 <a href="conform.html">conformance requirements</a>.</p>
 <p class="ready-for-wider-review">
-As in HTML [<a href="refs.html#ref-html51">HTML</a>, 10.4.2],
+As in HTML [<a href="refs.html#ref-html">HTML</a>, 10.4.2],
 all animated images with the same absolute URL and the same image
 data are expected to be rendered synchronised to the same timeline as a group,
 with the timeline starting at the time of the least recent addition to the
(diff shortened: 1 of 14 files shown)


svgeesus:
(do not merge yet, more commits incoming)

svgeesus:
ready for review

AmeliaBR:
Haven't reviewed yet, but: I think this might be a good use case for the "squash & merge" feature.

svgeesus:
Probably :) I thought small, single-purpose commits with meaningful names might be easier to check though

svgeesus:
Any other takers for a quick review?

svgeesus:
I think it is better to merge this now because other PR in the queue are likely to clobber it due to lots of text being moved around. The edits on this PR are very small and localized (updating hrefs from HTML5 to HTML LS)

master/render.html
```diff
@@ -638,7 +638,7 @@ <h3 id="PaintingRasterImages">Painting raster images</h3>
 requirements are discussed under
 <a href="conform.html">conformance requirements</a>.</p>
 <p class="ready-for-wider-review">
-As in HTML [<a href="refs.html#ref-html51">HTML</a>, 10.4.2],
+As in HTML [<a href="refs.html#ref-html">HTML</a>, 10.4.2],
```
AmeliaBR: With a belated review comment: that section number is wrong. Haven't tracked down the correct one.

If you have a chance, it could be fixed as a separate commit.

Issue #520, opened 2018-08-09
Update/move the normative-but-outdated SVG DOM appendix


AmeliaBR:
As I'm reviewing & tidying the changes appendix, I discovered we still have a normative ["SVG DOM" appendix](https://svgwg.org/svg2-draft/svgdom.html), which is in need of updates.

Required fixes include:

- Updating cross-references (e.g., to CSSOM, not DOM2 CSS).
- Clarify which parts are informative summaries of content normatively defined elsewhere.
- Redefine the section on initializing properties in a WebIDL consistent way (or remove, if the WebIDL definitions elsewhere already cover this).
- Check whether some of the specific behaviors actually match implementations:
  * Manually dispatched events on elements that don't normally receive events (I don't think we actually disallow event attributes on non-rendered elements, do we?)
  * Invalid values behavior

I suggest also:

- Merge any remaining content into the [Scripting and Interactivity chapter](https://svgwg.org/svg2-draft/interact.html).

I don't like having normative appendices, and one reason is they get overlooked and out of date!  Once we fix this one, the only other normative appendix is [Implementation Requirements](https://svgwg.org/svg2-draft/implnote.html). I'd prefer to move _its_ content into the relevant chapters, too, but at least it had a very "Hi, I'm Normative" nametag.

svgeesus:
The Implementation Requirements appendix started as somewhere convenient to park the lengthy mathematical derivation of the elliptical arc decomposition, which was done by Phillip Mansfield so that the last point of the command could be the current point of the next command (same as all the other path commands).

I agree that some of the other parts are easy to overlook. At minimum they should be well crosslinked from other parts of the spec; it might be a good idea to move some parts to the main spec. For example [Clamping Values](https://www.w3.org/TR/SVG2/implnote.html#RangeClamping) _might_ usefully move to [Basic Data Types](https://www.w3.org/TR/SVG2/types.html)?

AmeliaBR:
>  somewhere convenient to park the lengthy mathematical derivation of the elliptical arc decomposition

I agree that the conversion algorithms are appropriate for an appendix, because they aren't normative: they're just helpful notes for implementers. Similarly, the new "Notes on generating high-precision geometry" section is informative. They could stay in the appendix and it could be renamed "Implementation Notes".

For the rest:

- Error processing could go in [Conformance Criteria](https://svgwg.org/svg2-draft/conform.html).

- I agree, that Clamping Values makes sense in Basic Data Types.

[...]

PR #548, opened 2018-09-20
No normative appendices!

AmeliaBR:
Following the discussion in #520, this PR:

- Moves all normative content from the Implementation Requirements appendix into chapters.
- Renames the remainder of that appendix as non-normative "Implementation Notes"
- Updates all the normative content in the SVG DOM appendix.
- Merges that content into the Basic Data Types and the Scripting & Interactivity chapters.
- Leaves the svgdom.html file behind as a stub.

This is a lot of change, and some of it is important normative content!

I highly recommend reviewing it commit by commit, so you can separate the updates from the relocations.  Unfortunately, git is not very good at tracking the same content moved to different places.

Before merging, the changes entries need to be updated to include links to this PR.
There may also need to be some careful merging with PR #544 to not miss anything that's moved around.


master/svgdom.html
```diff
@@ -110,42 +110,57 @@ <h2 id="SVGDOMOverview">SVG DOM overview</h2>
   </li>
 </ul>
 
-<p>All SVG DOM objects that directly correspond to an attribute, e.g.
+<p>Many SVG DOM properties <a>reflect</a> a content attribute
```
svgeesus: Much nicer wording here

master/svgdom.html
```diff
@@ -110,42 +110,57 @@ <h2 id="SVGDOMOverview">SVG DOM overview</h2>
...
-  <dd>Initialized with the return-value of <a>SVGTextContentElement::getComputedTextLength</a>
-  on the same element.</dd>
-
   <dt class="DOMInterfaceName">DOMString</dt>
   <dd>Initialized as the empty string (<span class="attr-value">""</span>).</dd>
 
   <dt class="DOMInterfaceName">float</dt>
   <dt class="DOMInterfaceName">long</dt>
   <dt class="DOMInterfaceName">short</dt>
+  <dt>Any other <a href="https://heycam.github.io/webidl/#dfn-numeric-type">numeric type</a> defined in WebIDL</dt>
```
svgeesus: Hmm, yes

master/changes.html
```diff
@@ -1007,12 +1007,17 @@ <h3 id="svgdom">SVG Document Object Model (DOM)</h3>
...
+      "must" handle as browsers currently do (as SVGElement instances);
+      "should" implement SVGUnknownElement interface (though that is at risk).
+    </li>
     <li>Remove the special rule for initializing <a>SVGTextContentElement::textLength</a>,
       in favor of a clearer definition of the attribute's dynamic initial value.
     </li>
     <li>Clarify that rules about initializing objects
       only apply when reflecting <code>(none)</code> initial values,
-      and apply to interfaces no longer defined in this specification.
+      and also apply to interfaces defined in other specifications (e.g., Geometry).
```
svgeesus: much better

master/refs.html
```diff
@@ -19,15 +19,15 @@ <h1>References</h1>
 <h2 id="NormativeReferences">Normative references</h2>
 
 <dl class="references">
-  <dt id="ref-graphics-aria-1.0" class="normref">[<a href="https://www.w3.org/TR/graphics-aria-1.0/">graphics-aria-1.0</a>]</dt>
```
svgeesus: did this get deleted, or moved to informative?

svgeesus: ah okay moved, nevermind

master/svgdom.html
AmeliaBR: Yeah, that was added because the geometry interfaces (DOMPoint, DOMMatrix, etc.) are now spec'd to use doubles, not floats. And I wasn't going to list out all the variations currently in WebIDL!

master/changes.html
AmeliaBR: My own bad wording from the previous commit. Read it a second time & that's not quite what I meant to say!

master/types.html
```diff
@@ -207,6 +207,477 @@ <h2 id="Precision">Real number precision</h2>
...
+
+<p>The SVG DOM builds upon a number of DOM specifications.  In particular:</p>
+
+<ul>
+  <li>The SVG DOM requires, at a minimum, complete support for
+  <a href="https://www.w3.org/TR/dom/">DOM4</a>;
+  SVG software should implement
+  the <a href="https://dom.spec.whatwg.org/">latest DOM standard</a> 
+  wherever possible
+  [<a href="refs.html#ref-dom">DOM</a>]</li>
```
AmeliaBR: @svgeesus said:

> Removing W3C DOM4 from refs was a deliberate change. We should refer to DOM LS instead.

That sounds reasonable to me, but we can't really say they "must" implement the absolute latest spec, since there's often a lag between stabilizing a new spec feature & getting an implementation out.

Any suggestions for handling it better? "must support all aspects that were included in the DOM standard as of the publication of this specification and which haven't been deprecated or removed since"?

master/refs.html
AmeliaBR: Yeah, there were a few refs entries that were out of order. Probably should have made it a separate commit, but it's hard to add new entries to an alphabetical list if the starting list isn't sorted!

master/paths.html
```diff
@@ -1006,78 +1010,138 @@ <h2 id="PathDirectionality">Path directionality</h2>
...
-      "moveto" commands follow each other.</li>
-
-      <li>As mentioned in <a href="painting.html#StrokeProperties">Stroke Properties</a>,
-      linecaps must be painted for zero length subpaths when
-      <a>'stroke-linecap'</a> has a value of
-      <span class='prop-value'>round</span> or
-      <span class='prop-value'>square</span>.</li>
-    </ul>
+    <p>If either <strong>rx</strong> or <strong>ry</strong>
+    have negative signs, these are dropped;
```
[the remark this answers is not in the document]
AmeliaBR: Yeah, not how I would word it (I was just moving it around), but still seemed clear enough.

master/paths.html
```diff
@@ -1006,78 +1010,138 @@ <h2 id="PathDirectionality">Path directionality</h2>
...
+    a compound path data command, such as a "lineto" with
+    several pairs of coordinates. For example, for the path
+    data string <span class='attr-value'>'M 10,10 L 20,20,30'</span>,
+    there is an odd number of parameters for the "L" command, which requires an even
+    number of parameters. The user agent is required to draw
+    the line from (10,10) to (20,20) and then perform error
+    reporting since <span class='attr-value'>'L 20 20'</span>
+    is the last correctly defined segment of the path data specification.</li>
+
+    <li>Wherever possible, all SVG user agents shall report
```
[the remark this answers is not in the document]
AmeliaBR: Agree, it might be neater to downgrade to a "should" since the "wherever possible" already gives wiggle room.

In the general error processing guidelines (moved to conform.html lines 730+ in the final version of this PR), there's a longer bit about reporting errors to the user, but it is even more vague:

<blockquote>
<li>If the user agent has access to an error reporting
  capability such as status bar or console, it is recommended that the
  user agent provide whatever additional detail it can to
  enable the user or developer to quickly find the source of
  the error. For example, the user agent might provide an error
  message along with a line number and character number at
  which the error was encountered.</li>
</blockquote>

I added in the "or console" bit, otherwise the wording is straight from SVG 1.0.  As is the ["wherever possible, [UAs] shall" wording](https://www.w3.org/TR/SVG10/implnote.html#PathElementImplementationNotes).

AmeliaBR:
Thanks for the comments, Chris.

Definitely looking for reviews not only of my changes (& making sure I haven't dropped anything), but also of whether additional updates are required to some of the older text that has been lurking in the appendices all this time.

svgeesus:
@AmeliaBR I just squash-merged https://github.com/w3c/svgwg/pull/544 so you may need to rebase your branch.

master/types.html
svgeesus: I don't have any suggestions for handling it better. Saying that implementers must implement something that has been dropped from LS will make us unpopular.

master/types.html
```diff
@@ -207,6 +207,477 @@ <h2 id="Precision">Real number precision</h2>
...
+  <a href="https://www.w3.org/TR/uievents/">UI Events</a> and
+  <a href="https://www.w3.org/TR/clipboard-apis/">Clipboard API and events</a>
+  ([<a href="refs.html#ref-uievents">uievents</a>], [<a href="refs.html#ref-clipboard-apis">clipboard-apis</a>]).
+  (For the specific features that are required, see
+  see <a href="interact.html#RelationshipWithUIEVENTS">Relationship with UI Events</a>.)</li>
+
+  <li>The SVG DOM requires, at a minimum, complete
+  support for all interfaces from
+  <a href="https://www.w3.org/TR/DOM-Level-2-Style/Overview.html">DOM Level 2 Style</a>
+  ([<a href="refs.html#ref-dom-level-2-style">dom-level-2-style</a>])
```
svgeesus: We should really not be referring to DOM 2 Style or DOM 2 Views which are both ancient specifications. Views is [defined as obsoleted by HTML5](https://www.w3.org/TR/DOM-Level-2-Views/).

AmeliaBR: Yep, that's why I removed DOM Views completely.

Problem with DOM 2 Style is that _some_ of it is still relevant & required (while some is obsolete, never implemented), but it's replacement, CSS OM, is still only WD status & includes some new features which may not be stable. Which is why I went for the approach of requiring the union of the two specs: 

>  <li>The SVG DOM requires, at a minimum, complete
>   support for all interfaces from
>   <a href="https://www.w3.org/TR/DOM-Level-2-Style/Overview.html">DOM Level 2 Style</a>
>   ([<a href="refs.html#ref-dom-level-2-style">dom-level-2-style</a>])
>   that have not been dropped in the
>   <a href="https://drafts.csswg.org/cssom/">CSS Object Model (CSSOM)</a> specification.
>   SVG software should implement the latest version of
>   <a href="https://drafts.csswg.org/cssom/">CSS Object Model (CSSOM)</a>
>   wherever possible
>   ([<a href="refs.html#ref-cssom-1">cssom-1</a>])
>   </li>

master/types.html
AmeliaBR: Ok. That sort of matches how I'm handling the instability of CSSOM, anyway.

master/types.html
svgeesus: Fair enough, I agree with your reasoning

PR #549, opened 2018-09-21
Fix #532 with informative refs to ICC  & ICC support in css-color-4

### LINKED ISSUE #532: ICC support in Conforming High-Quality SVG viewer
[Conforming High-Quality SVG Viewer](https://www.w3.org/TR/SVG2/conform.html#ConformingHighQualitySVGViewers) says:

> Color management via ICC profile support (i.e., the ability to support colors defined using ICC profiles).

This has been in SVG since 1.0, is implemented in Inkscape, used to be implemented in the Adobe and Corel viewers. That functionality (`@profile` etc) is now in [CSS Color 4](https://drafts.csswg.org/css-color-4/#icc-colors) while SVG2 references CSS Color 3.

Using color profiles is a good and useful feature and should be encouraged. Conforming High-Quality SVG Viewer is in any case a clearly optional feature, so doesn't impede progress to Recommendation. So my preferred option would be to keep the wording, and add an informative reference to CSS Color 4 in references.

Thoughts?

svgeesus:
As this is an optional feature, informative references are fine. CSS Color 4 defines how to use ICC profiles in CSS, while the ICC ref covers the actual ICC profile format and meaning.

### DIFF
--- a/master/changes.html
+++ b/master/changes.html
@@ -135,6 +135,14 @@ have been made.</p>
     high-precision graphics to the Implementation Requirements appendix.
   </li>
 </ul>
+<div class='changed-since-cr1 cr2'>
+  <ul>
+    <li>Add informative references to ICC and to CSS Color 4
+      <a href="https://github.com/w3c/svgwg/issues/532">Issue discussion</a>
+      <a href="">Edits</a>
+    </li>
+  </ul>
+</div>
 
 <h3 id="rendering">Rendering Model chapter</h3>
 <ul>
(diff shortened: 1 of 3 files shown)


AmeliaBR:
You're missing edits to refs.html (adding the specs to the informative references section, sorted alphabetically by shortcode). Then make the [shortcodes] links to the reference entry.

Yes, I know, if we were using ReSpec or Bikeshed this would all be automatic, but we do things old-school here in the SVG WG.

You can copy & paste the markup from specref.org  Just be sure to remove the `<mark>` elements it adds around your search terms.

svgeesus:
blush

Yes, I am to used to bikeshed doing all that stuff automatically.

I will commit the refs (but possibly not today).

svgeesus:
OK, refs committed.

Issue #562, opened 2018-10-25
Syntactic forms: URL and <url> section is unclear or even wrong


dirkschulze:
Syntactic forms: URL and <url> does not differ correctly between URL and <url> (one being a "string", latter being a CSS Value/CSS function).

> However, many presentation attributes allow both URLs and text strings as content.
As far as I know, none of the presentation attributes use URL at all, they always use `url()` (or <url> CSS value).

Therefore, the next string:
> To disambiguate a text string from a relative URL
doesn't make sense. First, it is defined by the properties presentation attributes refer to anyway. Second, This described problem does not exist.

IMO, this section is misleading and could be removed entirely. If we want to make clear that there is a difference between URL and <url> we could add a non-normative note that presentation attributes always use the CSS Value <url> which is a CSS defined url() function and we link to CSS Units and Values and that <url> shall not be confused with URL which is a string as defined by the previous section.

@AmeliaBR ?

AmeliaBR:
The wording could definitely be clarified. As you say, it mostly comes down to "when you need a URL in a presentation attribute, it uses the CSS `url()` format". The main relevance in this chapter is that both types of URLs are covered by the rest of the section.

I agree that it is confusing to use the CSS type notation `<url>` without spelling out what that means (without following the link).

If you want to draft improved text, I'd be happy to review it.

PR #565, opened 2018-10-27
Make it more clear what the difference is between URL and <url>

dirkschulze:
Fix for #562

### DIFF
--- a/master/linking.html
+++ b/master/linking.html
@@ -181,7 +181,7 @@ class="attr-name">'href'</span></a> attribute.
 may be used for <a>presentation attributes</a> and their corresponding CSS properties
 [<a href="refs.html#ref-css-values-3">css-values</a>].</p>
 
-<p><a>URL</a> must not be a valid value for <a>presentation attributes</a> for structural
+<p><a>URL</a> is not a valid value for <a>presentation attributes</a> for structural
 relationships purposes. No non-<a>presentation attribute</a> allows <a>&lt;url&gt;</a> as value.</p>
 
 <p>SVG makes extensive use of <a>URL</a> references, both absolute and relative,
--- a/master/linking.html
+++ b/master/linking.html
@@ -164,7 +164,17 @@ escaping procedure and to refer to <a>URL</a> directly.</p>
 are specified using a <a>URL</a> value in an <a href="#linkRefAttrs"><span
 class="attr-name">'href'</span></a> attribute.</p>
 
-<p>This specification uses the terms <a>URL</a> and <a>&lt;url&gt;</a>.</p>
+<p>To describe linking relationships,
+this specification uses two different data types in attribute and property values:
+<a>URL</a> and <a>&lt;url&gt;</a>.
+The linking guidelines in this chapter apply to URLs specified with either syntax.
+</p>
+
+<p>The <a>URL</a> data type is a simple URL string.
+In SVG, most structural relationships between two elements
+are specified using a <a>URL</a> value in an <a href="#linkRefAttrs"><span
+class="attr-name">'href'</span></a> attribute.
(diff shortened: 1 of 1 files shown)


master/linking.html
```diff
@@ -161,14 +161,18 @@ <h3 id="URLandURI">URLs and URIs</h3>
...
-However, many <a>presentation attributes</a> allow both URLs and text strings as content.
-To disambiguate a text string from a relative URL, the <a>&lt;url&gt;</a>
-production is used for presentation attributes,
-and their corresponding CSS properties [<a href="refs.html#ref-css-values-3">css-values</a>].
-This is simply a URL delimited with a functional notation.
-</p>
+are specified using a <a>URL</a> value in an <a href="#linkRefAttrs"><span
+class="attr-name">'href'</span></a> attribute.</p>
+
+<p>This specification uses the terms <a>URL</a> and <a>&lt;url&gt;</a>.</p>
```
AmeliaBR: Suggestion for the first two paragraphs, swapping the order & adding some clarification (note, I'm not sure if "apply suggestion" will work here; it only seems to work for one line at a time).

```suggestion
<p>To describe linking relationships,
this specification uses two different data types in attribute and property values:
<a>URL</a> and <a>&lt;url&gt;</a>.
The linking guidelines in this chapter apply to URLs specified with either syntax.
</p>

<p>The <a>URL</a> data type is a simple URL string.
In SVG, most structural relationships between two elements
are specified using a <a>URL</a> value in an <a href="#linkRefAttrs"><span
class="attr-name">'href'</span></a> attribute.
</p>
```

(and then continue with the paragraph about the `<url>` data type)

svgeesus: @dirkschulze please accept (beta) the suggested change from @AmeliaBR and reply about the target of "must not"?

master/linking.html
```diff
@@ -161,14 +161,18 @@ <h3 id="URLandURI">URLs and URIs</h3>
...
+
+<p><a>&lt;url&gt;</a> is different from <a>URL</a> and represents a CSS <code>url()</code> function value.
+(See CSS Values and Units for further details [<a href="refs.html#ref-css-values-3">css-values</a>]). <a>&lt;url&gt;</a>s
+may be used for <a>presentation attributes</a> and their corresponding CSS properties
+[<a href="refs.html#ref-css-values-3">css-values</a>].</p>
+
+<p><a>URL</a> must not be a valid value for <a>presentation attributes</a> for structural
+relationships purposes. No non-<a>presentation attribute</a> allows <a>&lt;url&gt;</a> as value.</p>
```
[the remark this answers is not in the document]
svgeesus: Maybe "is not" ?IOW not a conformance clause, just an observation.

master/linking.html
```diff
@@ -161,14 +161,18 @@ <h3 id="URLandURI">URLs and URIs</h3>
...
+class="attr-name">'href'</span></a> attribute.</p>
+
+<p>This specification uses the terms <a>URL</a> and <a>&lt;url&gt;</a>.</p>
+
+<p><a>&lt;url&gt;</a> is different from <a>URL</a> and represents a CSS <code>url()</code> function value.
+(See CSS Values and Units for further details [<a href="refs.html#ref-css-values-3">css-values</a>]). <a>&lt;url&gt;</a>s
+may be used for <a>presentation attributes</a> and their corresponding CSS properties
+[<a href="refs.html#ref-css-values-3">css-values</a>].</p>
+
+<p><a>URL</a> must not be a valid value for <a>presentation attributes</a> for structural
```
svgeesus:
```suggestion
<p><a>URL</a> is not a valid value for <a>presentation attributes</a> for structural
```

Issue #560, opened 2018-10-19
<circle> with r=0 - disabled or not?


fsoder:
In SVG 1.1 r=0 on a `'circle'` element disabled rendering of the circle [1], while in SVG 2 this requirement has been removed [2]. I have not managed to find a paper trail for this behavioral change, and a quick UA sample did not show that the SVG 2 behavior has been implemented.
If this was an intentional change I wonder why it was only applied for `'circle'` and not the other shapes (`'rect'`, `'ellipse'`) which could reasonably render something even when degenerate.

[1] https://www.w3.org/TR/SVG11/shapes.html#CircleElementRAttribute
[2] https://svgwg.org/svg2-draft/shapes.html#CircleElement

AmeliaBR:
I don't think it was an intentional change. The equivalent statements about `rect` and `ellipse` are still there.

So, probably an accidental removal when the bits about `r` were moved to the geometry chapter.

tabatkins:
Agree this looks to be an accidental change.

I really wish we could kill this "degenerate shapes don't render" behavior; it makes such an annoying flicker when strokes or markers are involved...

EdorianDark:
Would it be possible to remove the "degenerated shapes don't render" behavior.

AmeliaBR:
> Would it be possible to remove the "degenerated shapes don't render" behavior.

We've discussed this previously, and came to the conclusion that it would likely not be backwards compatible.  Authors may be creating elements without width/height/radius, or with them set explicitly to zero, as a placeholder prior to updating them with script or SMIL animation.  How big of a compatibility issue it would be, I don't know. But I lean towards avoiding breaking changes whenever possible.

See #235 and #236 for some ideas and other complications.

dirkschulze:
Tests were added here: 
https://github.com/web-platform-tests/wpt/pull/13749#pullrequestreview-169079137

PR #566, opened 2018-10-27
Re-add text that would request to disable rendering an element if the…

dirkschulze:
… 'r' property/pres attribute computes to 0. #560

### DIFF
--- a/master/shapes.html
+++ b/master/shapes.html
@@ -201,7 +201,7 @@ radius.</p>
 
 <p>
   The <a>'r'</a> attribute defines the radius of the circle.
-  A negative value for either property is illegal and <a href="https://www.w3.org/TR/CSS21/syndata.html#parsing-errors">must be ignored as a parsing error</a>.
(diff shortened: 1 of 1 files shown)


master/shapes.html
```diff
@@ -201,7 +201,7 @@ <h2 id="CircleElement">The <span class="element-name">'circle'</span> element</h
 
 <p>
   The <a>'r'</a> attribute defines the radius of the circle.
-  A negative value for either property is illegal and <a href="https://www.w3.org/TR/CSS21/syndata.html#parsing-errors">must be ignored as a parsing error</a>.
+  A negative value for either property is illegal and <a href="https://www.w3.org/TR/CSS21/syndata.html#parsing-errors">must be ignored as a parsing error</a>. A computed value of zero disables rendering of the element.
```
AmeliaBR:
```suggestion
  A negative value is illegal and <a href="https://www.w3.org/TR/CSS21/syndata.html#parsing-errors">must be ignored as a parsing error</a>.
  A computed value of zero disables rendering of the element.
```

Fix another existing copy-paste problem with the text. (Remove "for either property" because we're only talking about one property here.)

Also: add a semantic line break before the new text.

PR #568, opened 2018-10-27
<circle> with r=0 has rendering disabled

### LINKED ISSUE #560: <circle> with r=0 - disabled or not?

ewilligers:
A computed radius value of zero disables rendering of the circle element.

This restores a sentence that was accidentally removed.

fixes #560

### DIFF
--- a/master/shapes.html
+++ b/master/shapes.html
@@ -201,7 +201,8 @@ radius.</p>
 
 <p>
   The <a>'r'</a> attribute defines the radius of the circle.
-  A negative value for either property is illegal and <a href="https://www.w3.org/TR/CSS21/syndata.html#parsing-errors">must be ignored as a parsing error</a>.
+  A negative value is illegal and <a href="https://www.w3.org/TR/CSS21/syndata.html#parsing-errors">must be ignored as a parsing error</a>.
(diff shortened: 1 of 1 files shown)


AmeliaBR:
I was very confused for a minute, wondering where my review disappeared to, but I'd added the review on @dirkschulze's version of the same change: https://github.com/w3c/svgwg/pull/566

Anyways, you caught the "for either property" bit, so my only suggestion is to put the new sentence on a new line.

Then whoever merges their PR first wins! (Or something like that…)

dirkschulze:
Let’s land this one

Issue #563, opened 2018-10-27
Defer HTML media elements in SVG 2.0


dirkschulze:
The definition of graphical elements include the elements audio, video, iframe and canvas

Should they get removed now?

https://svgwg.org/svg2-draft/struct.html#TermGraphicsElement

dirkschulze:
Closing this part of the issue. Rest gets tracked by #683.

Issue #448, opened 2018-05-14
Identify SVG1 elements that are obsolete in SVG2 


boggydigital:
Amelia: To slightly derail the conversation. This will be one of a number of obsolete elements from SVG 1. I think we may need to at least list them somewhere, for parsing purposes. E.g., HTML still recognizes these as valid SVG elements, even if they aren't implemented.

chris:I agree with Amelia that this needs to be done

AmeliaBR: Not against making it obsolete, just some clean-up work required.

Bogdan: Do we have a list of those elements?

Amelia: Not currently in the spec.

Bogdan: OK, so we'd need to compare with SVG 1 to see what has been removed.

Dirk: I wonder what Edge & other browsers currently do with these elements. Does the HTML parser already have special parsing rules for cursor element?

Amelia: That's a good question. Don't know the answer.

Bogdan: I know HTML spec says you shouldn't implement cursor, but I don't know beyond that.

Amelia: And there are other obsolete elements. Font-related elements, plus a few others.
https://oreillymedia.github.io/Using_SVG/guide/markup.html#obsolete

Dirk: Maybe best to discuss with the HTML editors, about what they think we should do.

Bogdan: I think we (Microsoft team) can take an action item to look into best ways forward.

boggydigital:
Not blocking updated 2.0 CR publication - assigning 2.0 Recommendation milestone to clean this up before 2.0 REC

svgeesus:
I think we need an obsolete elements, attributes, and attribute values chapter or appendix. 

Agenda+ to get some discussion.

svgeesus:
> Amelia: And there are other obsolete elements. Font-related elements, plus a few others.
https://oreillymedia.github.io/Using_SVG/guide/markup.html#obsolete

Nice list, though requiredFeatures and requiredExtensions should probably be added. Maybe xml:space too?

PR #571, opened 2018-10-28
Remove iframe, video, audio, canvas and unknown elements due to lack …

dirkschulze:
…of implementation

#563

### DIFF
--- a/master/changes.html
+++ b/master/changes.html
@@ -341,6 +341,9 @@ have been made.</p>
   <li>Removed the currentView and useCurrentView properties on <a>SVGSVGElement</a>.</li>
 
   <!-- <li>Added <a>SVGUnknownElement</a> for handling unknown elements in the svg namespace.</li> -->
+  <div class="changed-since-cr1 cr2 cr3">
+    <li>Removed SVGUnknownElement due to lack of implementations.</li>
+  </div>
 
   <li>Added the <a>'lang'</a> attribute in no namespace.</li>
 
@@ -688,11 +691,14 @@ have been made.</p>
 <h3 id="embedded">Embedded Content chapter (SVG 2 only)</h3>
 
 <ul>
-  <li>Added this chapter, which includes new
+  <div class="changed-since-cr1 cr2 cr3">
+    <li>Removed HTML multimedia section and the use of the audio, video, iframe, canvas elements due to lack of implementations.</li>
+  </div>
+  <!-- <li>Added this chapter, which includes new
   <a>'video'</a>, <a>'audio'</a>, <a>'iframe'</a>, <a>'canvas'</a>,
   <span class='element-name'>'source'</span> and <span class='element-name'>'track'</span>
   elements that behave almost identically
-  to the equivalent HTML elements.</li>
+  to the equivalent HTML elements.</li> -->
 
   <li>Removed the SVG copies of
   HTML's embedded content elements and instead allowed those
--- a/master/changes.html
+++ b/master/changes.html
@@ -340,7 +340,7 @@ have been made.</p>
 
   <li>Removed the currentView and useCurrentView properties on <a>SVGSVGElement</a>.</li>
 
-  <!-- <li>Added <a>SVGUnknownElement</a> for handling unknown elements in the svg namespace.</li> -->
+  <li>Added SVGUnknownElement for handling unknown elements in the svg namespace.</li>
   <div class="changed-since-cr1 cr2 cr3">
     <li>Removed SVGUnknownElement due to lack of implementations.</li>
   </div>
@@ -694,11 +694,11 @@ have been made.</p>
   <div class="changed-since-cr1 cr2 cr3">
     <li>Removed HTML multimedia section and the use of the audio, video, iframe, canvas elements due to lack of implementations.</li>
   </div>
-  <!-- <li>Added this chapter, which includes new
-  <a>'video'</a>, <a>'audio'</a>, <a>'iframe'</a>, <a>'canvas'</a>,
-  <span class='element-name'>'source'</span> and <span class='element-name'>'track'</span>
+  <li>Added this chapter, which includes new
+  'video', 'audio', 'iframe', 'canvas',
+  'source' and track'
   elements that behave almost identically
-  to the equivalent HTML elements.</li> -->
+  to the equivalent HTML elements.</li>
 
   <li>Removed the SVG copies of
   HTML's embedded content elements and instead allowed those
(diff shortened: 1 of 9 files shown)


master/changes.html
```diff
@@ -340,7 +340,10 @@ <h3 id="structure">Document Structure chapter</h3>
 
   <li>Removed the currentView and useCurrentView properties on <a>SVGSVGElement</a>.</li>
 
-  <li>Added <a>SVGUnknownElement</a> for handling unknown elements in the svg namespace.</li>
+  <!-- <li>Added <a>SVGUnknownElement</a> for handling unknown elements in the svg namespace.</li> -->
```
AmeliaBR:
```suggestion
  <li>Added SVGUnknownElement for handling unknown elements in the svg namespace (later removed).</li>
```

Keep the history in the changes section.

AmeliaBR: Similar comments for the rest of the edits to this file.

master/conform.html
```diff
@@ -368,28 +368,6 @@ <h2 id="referencing-modes">Processing modes for SVG sub-resource documents</h2>
...
-    <dt id="embedded-document-mode"><a>'iframe'</a> references</dt>
-    <dd>
-      <p>SVG documents referenced by the
-      an HTML <a>'iframe'</a> element in an SVG document must use the
-      same processing mode as the embedding document,
-      subject to any restrictions defined by the
-      <a href="https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-sandbox"><span class="attr-name">sandbox</span> attribute</a>
-      on the embedding <a>'iframe'</a>.</p>
-
-      <p class="note">The same processing rules are intended to be used when
```
AmeliaBR: We need to preserve the recommendations in this note, even without `<iframe>` in an SVG document.

master/embedded.html
```diff
@@ -15,31 +15,6 @@
 <body>
 
 <h1>Embedded Content</h1>
-<div class="annotation svg2-requirement">
```
AmeliaBR: Have we been removing annotations about deferred features? I would have expected to just add in information about the decision to defer.

master/struct.html
```diff
@@ -292,79 +292,6 @@ <h3 id="GElement">The <span class='element-name'>'g'</span> element</h3>
 
 <edit:elementsummary name='g'/>
 
-<h2 id="UnknownElement">Unknown elements</h2>
```
AmeliaBR: Instead of completely removing this section, there needs to be text to match browser behavior: unknown elements & their children must not be rendered, and must be represented in the DOM as generic `SVGElement` instances.

There also should be a mention about SVG 1 elements that are now obsolete (see #448). In particular, `<altGlyph>` should be treated as a `<tspan>` rather than disappearing.

master/types.html
```diff
@@ -359,18 +359,8 @@ <h3 id="ElementsInTheSVGDOM">Elements in the SVG DOM</h3>
 must also implement the DOM interface identified in the element definition.
 </p>
 </li>
-<li>
-<p>
-Elements in the SVG namespace whose local name does not match an element
-defined in any specification supported by the software
-should instead implement the <a>SVGUnknownElement</a> interface and
-must nonetheless implement the <a>SVGElement</a> interface.
```
AmeliaBR: The must part of this needs to be preserved! (It describes current browser behavior). Just delete line 366.

specs/animations/master/Overview.html
```diff
@@ -3428,7 +3428,7 @@ <h3 id="InterfaceSVGAnimationElement">Interface SVGAnimationElement</h3>
     <dl class="attributes">
       <dt id="__svg__SVGAnimationElement__targetElement" class="attribute first-child"><b>targetElement</b><span class="idl-type-parenthetical"> (readonly <a>SVGElement</a>?)</span></dt>
       <dd class="attribute">
-        <div>The element which is being animated. If no target element is being animated (for example, because the <a>'animate/href'</a> specifies an unknown element) the value returned is null.</div>
```
AmeliaBR: The "unknown" in this case refers to an unknown `id`, not an unknown tag name.  Maybe to make it clear, the parentheses could be

>  (for example, because the URL in <a>'animate/href'</a> does not match a valid element)

master/changes.html
dirkschulze: That is fine with me. Other specs more clearly differ between the individual TRs/CRs which makes it easier to follow the changes from a historical perspective.

master/conform.html
dirkschulze: We intended to add the annotation to make it clear for later edits why a certain feature was added or changed (maybe even removed) from the previous REC. In this case the feature weren't present in SVG 1.1 and dropped in SVG 2.0. Why do we want to keep the annotation?

master/embedded.html
dirkschulze: We keep those annotations in the 2.1 draft. So they are not lost. I still agree that we should keep annotation for moved, changed or added features?

master/conform.html
AmeliaBR: Processing modes in general are new in SVG 2. We have explicit requirements for processing modes for SVG embedded in other SVG, but we also have these notes with recommendations for processing modes for SVG embedded in HTML (which match what browsers, etc. do, but aren't explicitly spec'd anywhere).

Maybe there should be a separate sub-section, and the note about HTML images can be moved there, too?

svgeesus: Agree with Amelia that discussion of processing modes should be retained.

AmeliaBR:
Ping @dirkschulze. Are you able to follow-up on the comments/suggested changes this week?

svgeesus:
Looks like there is a little disagreement on how and where (2.0, 2.1) to document stuff that was added, and later removed.

For the unknown elements processing, maybe that should be split into a separate issue for more discussion, so we can merge this sooner?

master/changes.html
svgeesus: I made the similar edits in this section

master/struct.html
svgeesus: I didn't tackle this, it seems to need discussion on the text to add. My preference would be to raise that as a separate issue so we can land this PR.

> Instead of completely removing this section, there needs to be text to match browser behavior: unknown elements & their children must not be rendered, and must be represented in the DOM as generic SVGElement instances.

> There also should be a mention about SVG 1 elements that are now obsolete (see #448). In particular, <altGlyph> should be treated as a <tspan> rather than disappearing.

master/types.html
svgeesus: Kept (with the one-line deletion as suggested by @AmeliaBR )

specs/animations/master/Overview.html
svgeesus: Added

master/struct.html
longsonr: Firefox used to treat altGlyph as tspan, we removed all support for altGlyph some time ago though due to lack of use.

svgeesus: Now split out into a [separate issue](https://github.com/w3c/svgwg/issues/683)

master/conform.html
svgeesus: > Maybe there should be a separate sub-section, and the note about HTML images can be moved there, too?

In the interests of moving forward, I believe we should accept the PR and then add the separate subsection in a different PR.

AmeliaBR:
A reminder to please not merge significant PRs without running a test build. This left all sorts of broken cross-references.

I'm fixing it now, but for next time.

Issue #306, opened 2017-02-17
Since vector - effect options other than non - scaling - stroke already have multiple implementations, is the "at risk" flag valid?


satakagi:
Gecko patch: https://bugzilla.mozilla.org/show_bug.cgi?id=1318208
Polyfill: http://svg2.mbsrv.net/devinfo/devstd/non-scaling-objects-2/VectorEffect_polyfill_wScreen.html

jarek-foksa:
AFAIK `non-scaling-stroke` and "none" are the only values that have any implementations at all, thus the "at risk" flag seems to be justified.

nikosandronikos:
FWIW we were told that a polyfill is a valid implementation.

svgeesus:
Yes, a polyfill is a valid implementation for the purpose of of CR exit criteria.

tabatkins:
Tho, for web tech, polyfills aren't useful guarantors of quality/stability without additional assurances from browsers.

JonFerraiolo:
My opinion: drop everything but non scaling stroke. Lighten the burden on user agents, authoring tools, converters, book authors, students, and the WG itself. Keep it simple. In fact, I would not cry if the entire feature were dropped. If this set of functionality can't be achieved easily in Javascript, then better to add better DOM APIs into the spec and/or write a how-to blog post than add automagic features used only a tiny fraction of language users. 

Note that I have had a requirement for a non scaling stroke with SVG1 and found that I could achieve the desired result if I organized my transformations in the right way. It was not difficult.

satakagi:
@Tavmjong 
This is a test involving polyfill for that.
http://svgmap.org/devinfo/devstd/non-scaling-objects-2/VectorEffect_polyfill_noTf.html

However, I think that this does not conform to the requirements of WPT, so it is necessary to repair?

satakagi:
Made pull req. for tests 
https://github.com/web-platform-tests/wpt/pull/13302

dirkschulze:
We have implementations in Firefox, WebKit and Blink (Latter 2 counting as one implementation as implementation happened before fork.) That should fulfill requirements already but didn’t test Edge yet.

dirkschulze:
We have the other property values in the SVG 2.1 implementation. I suggest moving the other properties out of SVG 2.0.

longsonr:
[...]

PR #576, opened 2018-11-07
Remove at-risk flag from vector-effect values

dirkschulze:
#306


dirkschulze:
Apparently there was a misunderstanding on my side and the changes didn't land in Firefox yet. Please hold on with merging.

svgeesus:
As discussed on 3 Dec call, we changed our mind and want to leave vector effects as at-risk, so this can be closed.

Issue #326, opened 2017-06-15
SVGTransform's setMatrix method should take a DOMMatrix2DInit


zcorpan:
https://svgwg.org/svg2-draft/coords.html#InterfaceSVGTransform

```
void setMatrix(DOMMatrixReadOnly matrix);
```

This should be

```
void setMatrix(optional DOMMatrixInit matrix);
```

The prose can then invoke https://drafts.fxtf.org/geometry/#create-a-dommatrix-from-the-dictionary if you want an actual `DOMMatrix` to work with, but working with a dictionary works as well. In the latter case, https://drafts.fxtf.org/geometry/#matrix-validate-and-fixup should be invoked on the dictionary first.

zcorpan:
Same thing for https://svgwg.org/svg2-draft/coords.html#__svg__SVGTransformList__createSVGTransformFromMatrix

dstorey:
@dirkschulze this is relate to your recent PR #414. Can you handle this one too?

dirkschulze:
@dstorey Thought it is a similar approach, I don't think that the previous resolution covers DOMMatrixInit as well. Though I do support the change!

@zcorpan Why the change to `optional`?

zcorpan:
@dirkschulze WebIDL dictionaries must be optional.

> If the type of an argument is a dictionary type or a union type that has a dictionary as one of its flattened member types, and that dictionary type and its ancestors have no required members, and the argument is either the final argument or is followed only by optional arguments, then the argument must be specified as optional. Such arguments are always considered to have a default value of an empty dictionary, unless otherwise specified.

https://heycam.github.io/webidl/#ref-for-idl-dictionary%E2%91%A1

boggydigital:
Not blocking updated 2.0 CR publication - assigning 2.0 Recommendation milestone to clean this up before 2.0 REC

longsonr:
Can we confirm that...

a) I pass { a: 1, m11: 2 } we should throw a TypeError.
b) I pass {a: Infinity } and given that DOMMatrix2DInit takes unrestricted doubles unlike an SVGMatrix that should also result in a TypeError.

zcorpan:
Why was this specced to use `DOMMatrix2DInit` and not `DOMMatrixInit` (as suggested in OP)?

[...]

PR #577, opened 2018-11-07
Use DOMMatrix2DInit as parameters for methods. #326

dirkschulze:


master/coords.html
```diff
@@ -3226,7 +3226,7 @@ <h3 id="InterfaceSVGTransform">Interface SVGTransform</h3>
   [SameObject] readonly attribute <a>DOMMatrix</a> <a href="coords.html#__svg__SVGTransform__matrix">matrix</a>;
   readonly attribute float <a href="coords.html#__svg__SVGTransform__angle">angle</a>;
 
-  void <a href="coords.html#__svg__SVGTransform__setMatrix">setMatrix</a>(<a>DOMMatrixReadOnly</a> matrix);
+  void <a href="coords.html#__svg__SVGTransform__setMatrix">setMatrix</a>(optional <a>DOMMatrix2DInit</a> matrix);
```
AmeliaBR: Is the default value defined somewhere? Does a missing dictionary default to all zeros, or to an identity matrix? 2D or 3D?

dirkschulze:
@AmeliaBR ping?

AmeliaBR:
Sorry! Reviewing the linked geometry interfaces text, I think it safely covers the default behavior.

However, the way you have written the algorithm, I don't think you include the full error handling behavior from the validate and fix-up method, because you only run the steps as part of a hypothetical test for `is2D()`.

I would rewrite the steps as follows:

<ol class='algorithm'>
  <li>If the <a>SVGTransform</a> object is <a href='#ReadOnlyTransform'>read only</a>, then
  <a>throw</a> a <a>NoModificationAllowedError</a>.</li>
  <li> Let <var>newMatrix</var> be the result of <code><a>DOMMatrixReadOnly</a>.fromMatrix(<var>matrix</var>)</code>,
  including the <a href="https://drafts.fxtf.org/geometry/#dommatrixinit-dictionary">validate and fix-up</a> steps for missing values.
  If that method throws an error, then re-throw that error and abort these steps.</li>
  <li>If <var>newMatrix</var>.is2D() would return true,
  then set the <a>SVGTransform</a> object's
  <a href='#TransformValue'>value</a> to a <span class='prop-value'>matrix(…)</span>
  value that represents the same matrix as <var>newMatrix</var>.</li>
  <li>Otherwise, set the <a>SVGTransform</a> object's
  <a href='#TransformValue'>value</a> to a <span class='prop-value'>matrix3d(…)</span>
  value that represents the same matrix as <var>newMatrix</var>.</li>
  <li>In either case, set <var>newMatrix</var>
  as the <a>SVGTransform</a> object's <a href="#TransformMatrixObject">matrix object</a>.</li>
  <li>If the <a>SVGTransform</a> object
  <a href="#TransformMode">reflects a presentation attribute value of an element</a>,
  then <a>reserialize</a> the reflected attribute.</li>
</ol>

```html
<ol class='algorithm'>
  <li>If the <a>SVGTransform</a> object is <a href='#ReadOnlyTransform'>read only</a>, then
  <a>throw</a> a <a>NoModificationAllowedError</a>.</li>
  <li> Let <var>newMatrix</var> be the result of <code><a>DOMMatrixReadOnly</a>.fromMatrix(<var>matrix</var>)</code>,
  including the <a href="https://drafts.fxtf.org/geometry/#dommatrixinit-dictionary">validate and fix-up</a> steps for missing values.
  If that method throws an error, then re-throw that error and abort these steps.</li>
  <li>If <var>newMatrix</var>.is2D() would return true,
  then set the <a>SVGTransform</a> object's
  <a href='#TransformValue'>value</a> to a <span class='prop-value'>matrix(…)</span>
  value that represents the same matrix as <var>newMatrix</var>.</li>
  <li>Otherwise, set the <a>SVGTransform</a> object's
  <a href='#TransformValue'>value</a> to a <span class='prop-value'>matrix3d(…)</span>
  value that represents the same matrix as <var>newMatrix</var>.</li>
  <li>In either case, set <var>newMatrix</var>
  as the <a>SVGTransform</a> object's <a href="#TransformMatrixObject">matrix object</a>.</li>
  <li>If the <a>SVGTransform</a> object
  <a href="#TransformMode">reflects a presentation attribute value of an element</a>,
  then <a>reserialize</a> the reflected attribute.</li>
</ol>
```

AmeliaBR:
@dirkschulze Ok, those points make sense, too. My main point was that we actually have to run the full validate & fix-up algorithm to adjust the input value before using it, and throw any errors it issues.

Can you take those ideas and come up with a better edit?

AmeliaBR:
Ah, realized one thing is missing. This is a normative change to the API, so there should be a changes appendix entry.

Issue #322, opened 2017-06-05
Remove un-implemented `pointer-events` "painted" rules for images


AmeliaBR:
The [`pointer-events` definition](https://www.w3.org/TR/SVG2/interact.html#PointerEventsProp) contains detailed rules for "raster images" (presumably, any image embedded with `<image>`).  A key feature is that an image should only be considered "painted" in areas where it doesn't have full transparency.

There are two problems with this:

- it provides access to per-pixel image data without any requirements for cross-origin security restrictions
- none of the browser engines (as far as I can tell) implements it this way

Based on my testing ([CodePen demo](https://codepen.io/AmeliaBR/pen/qjEKXe?editors=1100)) browsers treat an `<image>` element with transparent regions as interactive over its full rectangular dimensions, regardless of whether the image is same or cross-origin.

BigBadaboom:
> none of the browser engines (as far as I can tell) implements it this way

I suspect that this was mostly because the spec has a somewhat ambiguous reading:

> For raster images, hit-testing is either performed on a whole-image basis (i.e., the rectangular area for the image is one of the determinants for whether the image receives the event) or on a per-pixel basis (i.e., the alpha values for pixels under the pointer help determine whether the image receives the event):

I think the original author was probably referring to the fact that some values ("visiblePainted" and "painted") require pixel level hit testing, and some don't. The former options include the the text:

> if any pixel from the raster image which is under the pointer is not fully transparent

whereas the remaining options talk about:

> the rectangular area of the image

A loose reading of the introductory paragraph could be interpreted to mean that implementations had a choice whether to use pixels or not.  I hope the spec doesn't get changed to remove the per-pixel options, and if it doesn't, then perhaps the intro should be modified to remove any ambiguity.

Browser bug reports for this issue here:
- https://bugzilla.mozilla.org/show_bug.cgi?id=311942
- https://bugs.chromium.org/p/chromium/issues/detail?id=806468

BigBadaboom:
This issue has popped up on Stack Overflow a few times recently.  People expect transparent pixels to be transparent to clicks.

Currently, the only way to get the expected behaviour is to use a `<clipPath>`. But if they are not using pre-defined images, cliPaths are not really feasible.

https://stackoverflow.com/questions/48447142/svg-how-to-ignore-pointer-events-on-transparent-parts-of-an-image

boggydigital:
Not blocking updated 2.0 CR publication - assigning 2.1 WD milestone

[...]

PR #584, opened 2018-11-09
CSS pointer-events on (raster) images does not require scanning pixels

dirkschulze:
…requirement to scan raster image pixels. #322

### DIFF
--- a/master/changes.html
+++ b/master/changes.html
@@ -883,7 +883,7 @@ have been made.</p>
   the 'cursor element' element to take an URL that is not in a CSS-like
   functional form.</li>
   <li>Added the <span class="prop-value">bounding-box</span> keyword to <a>'pointer-events'</a>.</li>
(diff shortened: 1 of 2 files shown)


master/changes.html
```diff
@@ -883,6 +883,9 @@ <h3 id="interact">Scripting and Interactivity chapter (separate chapters in SVG
   the 'cursor element' element to take an URL that is not in a CSS-like
   functional form.</li>
   <li>Added the <span class="prop-value">bounding-box</span> keyword to <a>'pointer-events'</a>.</li>
+  <div class='changed-since-cr1 cr2'>
```
svgeesus: Shouldn't that be

```
<div class='changed-since-cr1 cr2 cr3'>
```

dirkschulze: We indeed have 3 CRs already. Hm. There might be issues in other, already merged PRs as well. Will take a look this week.

Issue #578, opened 2018-11-07
Drawing vector-effects mentions SVG 1.2T terms


dirkschulze:
https://svgwg.org/svg2-draft/painting.html#PaintingVectorEffects

> More precisely: a user agent establishes a host coordinate space which in SVG Tiny 1.2 is always the same as "screen coordinate space".

I wonder if we could simply remove this sentence.

Issue #582, opened 2018-11-09
vector-effect host coordinate space is not "screen"


dirkschulze:
The spec text for rendering of `vector-effect: non-scaling-stroke` says:

>  With the non-scaling-stroke vector effect, stroke outline shall be calculated in the "host" coordinate space instead of user coordinate system. More precisely: a user agent establishes a host coordinate space which in SVG Tiny 1.2 is always the same as "screen coordinate space".

https://svgwg.org/svg2-draft/painting.html#PaintingVectorEffects

The definition of `vector-effect` in the spec has the following:

> **viewport**
> Specifies immediate viewport coordinate system as the host coordinate space. When that element belongs to nested viewport coordinate system, vector effects are applied toward viewport coordinate system to which that element belongs directly. That is, that vector effect is not effective for change of CTM on ancestral viewport coordinate system.
>
> **screen**
> It specifies the coordinate system of content which under the immediate control of user agent. So to speak, it is "scrren" which user agent has. ("screen coordinate space" in SVGT1.2) Even if that element belongs to nested viewport coordinate system, that vector effect is always effective for change of CTM of the any hierarchy.

`viewport` is the default:

> An initial value in case it is not specified is **viewport**.

https://svgwg.org/svg2-draft/coords.html#VectorEffects

Nothing of that is what browsers implement. I tested the following scenarios `<rect>` with `non-scaling-stroke`:
1. in nested, inner SVG element with `viewBox` scaling.
    https://codepen.io/krit/pen/KrMgWd?editors=1100
2. in inline SVG with CSS Transform on root-`<div>` box:
    https://codepen.io/krit/pen/RqRRgZ?editors=1100
3. in inline SVG within `<div>`-box within `<foreignObject>` in SVG document.
    https://codepen.io/krit/pen/xQOOBg?editors=1100
4. in 3D rendering context with 2D transform on group element establishing 3D rendering context.
    https://codepen.io/krit/pen/ZmOpaL?editors=1100 (only Firefox)

All examples have an overlaying `<rect>` which is not in a scaled coordinate space as reference. If you see red, then `non-scaling-stroke` does not apply to the same context as the reference `<rect>`.

**Findings:**
* The "host" coordinate system is the one on context change (SVG embedded in HTML context). Even inside a `<foreignObject>` the context change would be on a boundary to HTML. TODO: Test SVG only contexts within `<foreignObject>`
[...]

PR #585, opened 2018-11-10
Adapt host coordinate space def of vector-effect to implemented behav…

dirkschulze:
…ior. #582 

The definition is only affecting browsers or SVG UAs with support of other XML languages (or alike languages like HTML). For pure SVG UAs, the host coordinate space is the same as the initial coordinate space. Note that this might not be equivalent to the device coordinate space.

### DIFF
--- a/master/changes.html
+++ b/master/changes.html
@@ -525,6 +525,9 @@ have been made.</p>
   <li>Define SVG sizing in CSS contexts with reference to the CSS Default Sizing Algorithm.</li>
   <li>Clarify that percentages are relative to the width and height of the specified viewBox.</li>
   <li>Mark unimplemented <a>'vector-effect'</a> options at-risk.</li>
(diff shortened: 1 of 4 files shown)


master/changes.html
```diff
@@ -525,6 +525,9 @@ <h3 id="coords">Coordinate Systems, Transformations and Units chapter</h3>
   <li>Define SVG sizing in CSS contexts with reference to the CSS Default Sizing Algorithm.</li>
   <li>Clarify that percentages are relative to the width and height of the specified viewBox.</li>
   <li>Mark unimplemented <a>'vector-effect'</a> options at-risk.</li>
+  <div class='changed-since-cr1 cr2'>
```
svgeesus: cr1 cr2 cr3

dirkschulze:
Also fixes #578

Issue #555, opened 2018-10-13
Update property definitions to use new CSS approach to "animatable"


AmeliaBR:
See @fantasai's [email defining the change (*Member only archive*)](https://lists.w3.org/Archives/Member/w3c-css-wg/2018OctDec/0037.html) or the [public version outlining the changes](https://lists.w3.org/Archives/Public/www-style/2018Oct/0007.html).

svgeesus:
Yes, this will require an entire-spec editing pass on all properties.

This doesn't affect implementations, afaict, it is just a cleaner way to write the specification.

dirkschulze:
Done.

Issue #534, opened 2018-08-20
Review property value definitions to ensure unitless lengths are correctly included


AmeliaBR:
See discussion in #529, where it was discovered that `stroke-width` syntax definition in SVG 2 erroneously left out unitless length. There may be other errors.

ewilligers:
Lists of properties that accept a numeric value, in quirks mode and standard mode: https://docs.google.com/spreadsheets/d/1HRzb_-S28xq7GqYKhHbMP2YQlqWLvOWSS_twLXy-I40/edit?usp=sharing

ewilligers:
Raised https://github.com/w3c/csswg-drafts/issues/3057
Blink and Safari also support numeric values for `cx` `cy` `r` `rx` `ry` `x` `y`.

svgeesus:
[CSS WG discussed this and resolved](https://github.com/w3c/csswg-drafts/issues/3057#issuecomment-418918339)

- RESOLVED: Only accept unitless values on SVG 1.1 supported properties with the exception of baseline-shift

svgeesus:
So, is the `stroke-width` definition in SVG2 the normative one, or the one in CSS fill and stroke?

@ericwilligers looks like we need a bit on SVGWG discussion of [your document](https://docs.google.com/spreadsheets/d/1HRzb_-S28xq7GqYKhHbMP2YQlqWLvOWSS_twLXy-I40/edit#gid=0) before making edits to SVG2?

ewilligers:
`stroke-dashoffset` and `stroke-width` accept unitless lengths in SVG 1.1, and in all browsers.
The SVG 1.1 definition for [length](https://www.w3.org/TR/SVG11/types.html#DataTypeLength) says
> A length is a distance measurement, given as a number along with a unit which may be optional.

We kept the SVG 1.1 grammar `<length>` for these properties, but this is incorrect as we changed the definition of `<length>`. We need to change `stroke-dashoffset` and `stroke-width` to accept `<number>`


> is the stroke-width definition in SVG2 the normative one, or the one in CSS fill and stroke?

The SVG 2 definition is normative. CSS Fill and Stroke is not a normative reference for SVG 2.

dirkschulze:
I think I covered the changes. @ericwilligers please review otherwise I'll close the issue within the next days.

PR #586, opened 2018-11-10
Unify CSS properties, use new animation type attributes, allow number…

dirkschulze:
… values for old CSS properties, align computed styles, fix prop index. #534, #529, #555

### DIFF
--- a/master/definitions.xml
+++ b/master/definitions.xml
@@ -1219,6 +1219,7 @@
   <interface name='CanvasDrawPath' href='https://html.spec.whatwg.org/multipage/canvas.html#canvasdrawpath' />
 
   <!-- ... terms .......................................................... -->
+  <term name='Animation type' href='https://www.w3.org/TR/web-animations-1/#animating-properties'/>
   <term name='event type' href='https://dom.spec.whatwg.org/#ref-for-dom-event-type'/>
   <term name='event handler content attribute' href='https://html.spec.whatwg.org/multipage/webappapis.html#event-handler-content-attributes'/>
   <term name='compound selector' href='http://dev.w3.org/csswg/selectors4/#compound'/>
--- a/master/definitions.xml
+++ b/master/definitions.xml
@@ -1219,7 +1219,8 @@
   <interface name='CanvasDrawPath' href='https://html.spec.whatwg.org/multipage/canvas.html#canvasdrawpath' />
 
   <!-- ... terms .......................................................... -->
-  <term name='Animation type' href='https://www.w3.org/TR/web-animations-1/#animating-properties'/>
+  <term name='Animation type' href='https://www.w3.org/TR/web-animations-1/#animation-type'/>
+  <term name='not additive' href='https://drafts.csswg.org/css-values-4/#not-additive'/>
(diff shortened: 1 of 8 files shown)


dirkschulze:
@birtles Maybe you get the time to review the changes to CSS Properties “Animation type”?

master/painting.html
```diff
@@ -1222,14 +1224,45 @@ <h3 id="StrokeDashing">Dashing strokes: the <span class="property">'stroke-dasha
 <a>'stroke-dasharray'</a>: each dash and gap length is interpreted relative
 to the author's path length as specified by <a>'path/pathLength'</a>.</p>
 
+<p id="interpolationDashPattern"><a>'stroke-dasharray'</a> must use discrete
+interpolation if <var>start</var> or <var>end</var> value compute to <span class="prop-value">none</span>
+or are invalid.
+Otherwise, repeat both dash patterns of <var>start</var> and <var>end</var> value list until the length
+of elements in both value lists match. Each item is then combined by computed value.
+The interpolation is <a href="https://svgwg.org/specs/animations/#AdditionAttributes">non-additive</a>.</p>
```
birtles: We should link to CSS V&U 4 here: https://drafts.csswg.org/css-values-4/#not-additive

master/painting.html
```diff
@@ -1222,14 +1224,45 @@ <h3 id="StrokeDashing">Dashing strokes: the <span class="property">'stroke-dasha
 <a>'stroke-dasharray'</a>: each dash and gap length is interpreted relative
 to the author's path length as specified by <a>'path/pathLength'</a>.</p>
 
+<p id="interpolationDashPattern"><a>'stroke-dasharray'</a> must use discrete
+interpolation if <var>start</var> or <var>end</var> value compute to <span class="prop-value">none</span>
+or are invalid.
```
birtles: I haven't checked that this is what browsers actually do, but assuming it is. This looks good. I'd personally find the following a little easier to follow but it's up to you:

> 'stroke-dasharray' values are non-additive. For interpolation, 'stroke-dasharray' values are combined as follows:
> * If either |start| or |end| compute to 'none' or are invalid, they are combined using the discrete animation type.
> * Otherwise, repeat both dash patterns of |start| and |end| value list until the length of elements in both value lists match. Each item is then combined by computed value.

master/definitions.xml
```diff
@@ -1219,6 +1219,7 @@
   <interface name='CanvasDrawPath' href='https://html.spec.whatwg.org/multipage/canvas.html#canvasdrawpath' />
 
   <!-- ... terms .......................................................... -->
+  <term name='Animation type' href='https://www.w3.org/TR/web-animations-1/#animating-properties'/>
```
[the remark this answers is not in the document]
dirkschulze: Done. SVG build system doesn't make it so easy to switch between ED and WD depending on the publication type as Bikeshed. Since the spec is in CR going to PR it seems saver to use TR links.

master/painting.html
dirkschulze: done (even though it is ED :)). Added it to the linking section in definitions.xml for easy edit.

master/painting.html
dirkschulze: Sounds better. Changed it to that.

Issue #456, opened 2018-05-16
isPointInFill/isPointInStroke and display: none and other specifics


dirkschulze:
WebKit (in nightly) and Blink return `false` always for:

`isPointInFill`
* if `display: none`
* (if `fill: none`) only on Blink    **Edit:** Blink returns `true` now! Thanks @fsoder.
* the element is _not_ attached to the current document context

`isPointInStroke`
* if `display: none`
* if `stroke: none`   **Edit:** Blink returns `true` now! Thanks @fsoder.
* the element is _not_ attached to the current document context

Both, webkit and blink, would not be able to compute `isPointIn*` for `display: none` or not-attached elements. This is an implementation detail that both engines can't easily work around. I believe Firefox has the same issue. CC @longsonr.

For `stroke: none` and `fill: none` there probably is no implementation issue. At least on WebKit. We just need to decide if `none` values on `stroke` or `fill` will have an effect for the 2 methods.

Spec text: https://svgwg.org/svg2-draft/types.html#InterfaceSVGGeometryElement
Test case: https://codepen.io/krit/pen/VxEbdp

AmeliaBR:
The methods should return valid values even if the shape isn't currently stroked / filled, since we have `pointer-events` settings that allow the stroke/fill region to be pointer-sensitive in those situations.

I would prefer for the methods to return valid values even for non-rendered elements, just like `getBBox()` is supposed to do (and does in most browsers).

If implementations cannot do that, the methods should throw an error, not just silently return false.

dirkschulze:
@AmeliaBR even for getBBox() this is no longer the case in many situations in Blink or WebKit since both implemented the Geometry CSS properties. Without a render tree (which means attached to the tree and not display: none, the values can not get computed. I expect that they’ll return an empty box in Firefox as well once Gecko implements those properties. I’ll leave that to Robert to clarify.

dirkschulze:
Here 3 tests. Each test has 3 rectangles:
1. rect in `<defs>` section
2. rect with `display: none;`
3. normal rect for reference.

**1. `getBBox()`**
https://codepen.io/krit/pen/wQzPQj
* Chrome, WebKit, Edge: **1.** and **3.** show the correct bounding box. **2.** has an SVGRect with all values set to `0`.
* Firefox, `NS_ERROR` for **1.** and **2.**. **3.** shows the correct bounding box.

**2. `isPointInFill()`**
https://codepen.io/krit/pen/xQEpep
[...]

PR #594, opened 2018-11-19
Clarify possible UA behavior on getTotalLength, getPointAtLength, get…

dirkschulze:
…BBox, getCTM, getScreenCTM, isPointInFill and isPointInStroke if the current element is a non-rendered element. #456

### DIFF
--- a/master/changes.html
+++ b/master/changes.html
@@ -287,6 +287,12 @@ have been made.</p>
   </li>
 </ul>
 </div>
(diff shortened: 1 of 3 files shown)


ewilligers:
Should we also mention that detached elements are non-rendered elements? Perhaps the definition of "non-rendered element".

dirkschulze:
Created an issue for non-attached elements: https://github.com/w3c/svgwg/issues/599

Issue #595, opened 2018-11-21
Possible inconsistency with set element and href attribute


waruyama:
In the specifiation for the `set` element at https://svgwg.org/specs/animations/#SetElement there is no mentioning of the `href` attribute. 

However, in the SVG 2 spec, *Chapter 5: Document Structure* there is an example where the `href` attribute is used in on `set` element: https://www.w3.org/TR/SVG2/struct.html#UseAnimations

```
<set href="#target" begin="mouseover" ... />
  <!-- only affects the element that is moused over -->

<set href="#target" begin="target.mouseover" ... />
  <!-- affects all instances of the element with the id 'target',
          in all light and shadow node trees,
          when any of them are moused over -->
```

Note: In SVG 1.1 the `xlink` attributes were listed in the `set` element: https://www.w3.org/TR/SVG11/animate.html#SetElement

PR #597, opened 2018-11-21
Fix href in animation

birtles:
See #595.


birtles:
Actually, since I'm the only editor of this spec, I guess I can just merge it.

PR #598, opened 2018-11-22
Fixes #596

### LINKED ISSUE #596: Place of SVG presentation attributes in the CSS cascade
The current ED of SVG2 [says](https://svgwg.org/svg2-draft/styling.html#PresentationAttributes):

> Presentation attributes contribute to the author level of the cascade, _following all other_ author-level style sheets, and have specificity 0.

This seems to imply that style declarations "generated" from the presentation attributes are added _at the end_ of the author styles and thus should override the "regular" author styles with zero specificity (e.g. `* { fill: red }`). This is dramatically different from the definition in SVG 1.1, where they were inserted "_at the start_ of the author style sheet", and since browsers seem to follow the SVG 1.1 rules, the change would be unlikely web-compatible.

Isn't it a typo? Shouldn't the spec say "..._followed by_ all other author-level style sheets" instead (or even something more unambiguous, like that "...at the start of..." from the old spec)?

SelenIT:
Fixes the priority of the presentation attributes in the CSS cascade (reverting it to the SVG 1.1 definition, which matches the browser reality).

### DIFF
--- a/master/styling.html
+++ b/master/styling.html
@@ -305,10 +305,10 @@ text.error   { fill: red; }</pre>
 and <a>'style attribute'</a> attributes, but also in
 <dfn id="TermPresentationAttribute" data-dfn-type="dfn" data-export="">presentation attributes</dfn>.
 These are attributes whose name matches (or is similar to) a given CSS property
-and whose value is parsed as a value of that property.  Presentation
+and whose value is parsed as a value of that property. Presentation
 attributes contribute to the
 <a href="https://drafts.csswg.org/css-cascade-3/#preshint">author level</a>
-of the cascade, following all other author-level style sheets,
+of the cascade, followed by all other author-level style sheets,
 and have specificity 0.</p>
 
 <p>Since presentation attributes are parsed as CSS values, not declarations, an


ewilligers:
The SVG 2 wording remains confusing. By "followed by all other author-level style sheets", I suspect you mean to be consistent with the SVG 1.1 wording " the presentation attributes are conceptually inserted into a new author style sheet which is the first in the author style sheet collection. The presentation attributes thus will participate in the CSS2 cascade as if they were replaced by corresponding CSS style rules placed at the start of the author style sheet with a specificity of zero."

I suspect the existing wording "following all other author-level style sheets" means to refer to the order in
[Origin and Importance](https://drafts.csswg.org/css-cascade-3/#cascading), i.e. after "Normal author declarations" and before "Normal user declarations".

The SVG 1.1 wording was clearer. Why don't we go back to that?

ewilligers:

dirkschulze:
@heycam Can you summarize the reason of the change?

SelenIT:
@ewilligers, I don't see this ambiguity in the SVG2 wording. It clearly says that presentation attributes "contribute" to the Author styles, not "between" the two Origins. Also, AFAIK, normal Authors styles come after normal User styles in the cascade, so if presentstion attributes were

dirkschulze:
@ericwilligers IMO, the wording references CSS Cascading now and declares presentation attribute with the help of that spec. Sounds good to me.

PR #603, opened 2018-12-02
Add CSS properties cx, cy, r, rx, ry, x, y to property index. #82

dirkschulze:
Part 1, add geometry properties to prop index.


dirkschulze:
This was part 1 of #82

PR #612, opened 2018-12-08
Remove properties defined by other CSS modules from prperty index tab…

dirkschulze:
…le. #82

### DIFF
--- a/master/propidx.html
+++ b/master/propidx.html
@@ -36,45 +36,6 @@
       </thead>
       <tbody>
         <tr>
-          <th><a>'alignment-baseline'</a></th>
-          <td>auto | baseline | before-edge | text-before-edge |
-          middle | central | after-edge | text-after-edge |
-          ideographic | alphabetic | hanging | mathematical </td>
-          <td>see property description</td>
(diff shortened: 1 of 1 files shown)


ewilligers:
We may need to keep [image-rendering](https://svgwg.org/svg2-draft/painting.html#ImageRendering).

dirkschulze:
@ericwilligers Ah, no. This should be removed as well. Covered by CSS Images https://drafts.csswg.org/css-images-3/#the-image-rendering

ewilligers:
> Covered by CSS Images 

Only the Images ED has image-rendering. For now, SVG2 has the [normative definition](https://svgwg.org/svg2-draft/painting.html#ImageRendering).

dirkschulze:
@Tavmjong thought so too. @tabatkins what are the next steps on CSS Images 4?

ewilligers:
There is already some [browser support](https://wpt.fyi/results/css/css-images/parsing/image-rendering-valid.html?label=stable&aligned) for 'pixelated' and 'crisp-edges', but not 'smooth' or 'high-quality'.

Implementations can use the new values without SVG 2 having a normative dependency on Images 4.

Issue #427, opened 2018-04-30
Clarify hit testing behavior of <foreignObject>


chrishtr:
Should foreignObject be considered a hit testing target?
http://output.jsbin.com/garirac/quiet
Testing this (click to the left of the blue square) on Chrome 68 (tip of tree), Firefox and Edge seems to indicate the answer should be Yes. See also https://chromium.googlesource.com/chromium/src.git/+/b19cea8da334391dc1a51e2d386b10b25e354a69,
which recently changed this behavior in Chrome.

boggydigital:
Not blocking updated 2.0 CR publication - assigning 2.0 Recommendation milestone to clean this up before 2.0 REC

PR #613, opened 2018-12-09
Clarify that bounding box of foreignObject is hit-teasting target and…

dirkschulze:
… content may be too. #427

### DIFF
--- a/master/changes.html
+++ b/master/changes.html
@@ -170,7 +170,7 @@ have been made.</p>
   <li class='changed-since-cr1'>Deferred z-index support to a future version of SVG.
     <a href="https://github.com/w3c/svgwg/issues/483">Issue discussion</a>
     <a href="https://github.com/w3c/svgwg/pull/489">Edits</a></li>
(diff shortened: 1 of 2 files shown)


master/changes.html
```diff
@@ -167,22 +167,14 @@ <h3 id="rendering">Rendering Model chapter</h3>
...
+  <li class='changed-since-cr1'>Deferred z-index support to a future version of SVG.
     <a href="https://github.com/w3c/svgwg/issues/483">Issue discussion</a>
     <a href="https://github.com/w3c/svgwg/pull/489">Edits</a></li>
-</ul>
-</div>
-
-<div class='changed-since-cr1 cr2'>
-  <ul>
-    <li>Removed the use element from list of elements that the <a>'visibility'</a> property directly affects.
+  <li class='changed-since-cr1 cr3'>Clarify that <a>'foreignObject'</a> is and its content may be hit-testing target.</li>
```
svgeesus:
```suggestion
  <li class='changed-since-cr1 cr3'>Clarify that <a>'foreignObject'</a> is, and its content may be, hit-testing target.</li>
```

svgeesus: the subordinate clause is more readable with the commas.

Issue #211, opened 2016-07-19
Specify which attributes/properties are horizontal vs vertical vs other


AmeliaBR:
Most length attributes & properties have percentages that "refer to the size of the current SVG viewport" (see #209).  However, what that means depends on whether it is an x/width-based length, a y/height-based length, or other.

It would be preferable to define this distinction explicitly for each property or attribute.  In other words, replace  "refer to the size of the current SVG viewport" with "refer to the viewport width" or "refer to the viewport height" where appropriate.

nikosandronikos:
Ahh yes, sorry I forgot to reply to this point. I intended to say that I support the idea of stating explicitly per attribute what the percentage is relative to, but want to do so as a new issue and deal with it a bit later.

AmeliaBR:
I discovered a case of broken interop caused by the lack of clarity about how percentages are resolved: the `textLength` attribute.

Safari, Edge and Firefox treat this as an other/diagonal length.  (Although Safari has an unrelated bug where they add extra space after the last character.)

However, Chrome always resolves against width (even for vertical writing mode).

[Test case](https://codepen.io/AmeliaBR/pen/4cab473d71302f0f2b238e3dee1f5b1d/)

Probably fair to call this a Chrome bug, but it isn't explicitly defined in the spec, either.

fsoder:
When faced with the decision (due to an implementation inconsistency) I ended up picking 'width' - because of this being a static mapping in Blink, otherwise it'd have been 'width' / 'height'... maybe eventually we'll be able to fix that. I'm pretty sure we had either 'width' / 'height' or 'width' in Presto as well. Having a hard time seeing this as an 'other' length TBH.

BigBadaboom:
By "other" are you just meaning the diagonal one, or are you including other variants like those that use path length, for instance?

AmeliaBR:
> By "other" are you just meaning the diagonal one, or are you including other variants like those that use path length, for instance?

By "other", I'm talking specifically about the diagonal definition for percentages relative "size of the current SVG viewport".  For percentages relative to anything else, I think we have clear definitions in the specs of what they are measured against.  (If there are cases where we don't, that would be a separate issue.)

BigBadaboom:
If we are going to mark some attributes with their "percent method", then why not be consistent and mark up all relevant attributes the same way?

AmeliaBR:
> Having a hard time seeing this as an 'other' length TBH.

[...]

PR #615, opened 2018-12-10
Define which dimension percentage values of CSS properties are relet…

dirkschulze:
…ive to. #211


dirkschulze:
This only covers CSS properties. There needs to be a follow up on SVG attributes.

PR #620, opened 2018-12-27
<path d> parsing spec needs a special case to not use CSS syntax

SimonSapin:
When reading the spec literally before this changes, https://svgwg.org/svg2-draft/styling.html#PresentationAttributes specifies that all presentation properties are parsed according to https://svgwg.org/svg2-draft/types.html#presentation-attribute-css-value

For the `d` property, this means the CSS grammar `none | <string>`.

The following would be valid:

* `<path d="none">`
* `<path d="'M 100 100 L 300 100 L 200 300 z'">` (not the single quotes)
* `<path d="/**/'M 100 100 \L \33 00 100 L 200 300 z'">` (equivalent to the previous)

But `<path d="M 100 100 L 300 100 L 200 300 z">` or indeed any SVG 1.1 path would not be valid because they parse as CSS ident and number tokens, not as a `<string>`.

https://github.com/w3c/svgwg/issues/320 poposes changing the syntax of the `d` CSS propery to be even further to the 1.1 attribute syntax.

This does not appear to be an intentional change from SVG 1.1, so this pull request "reverts" it.

### DIFF
--- a/master/styling.html
+++ b/master/styling.html
@@ -314,8 +314,9 @@ and have specificity 0.</p>
 <p>Since presentation attributes are parsed as CSS values, not declarations, an
 <a href="https://www.w3.org/TR/2011/REC-CSS2-20110607/cascade.html#important-rules"><span class="prop-value">!important</span> declaration</a>
 within a presentation attribute will cause it to have an <a>invalid value</a>.
-See <a href="types.html#presentation-attribute-css-value">Attribute syntax</a>
-for details on how presentation attributes are parsed.</p>
+Except as noted in thetable for the <a>'d'</a> presentation attribute,
+the value of presentation attributes is parsed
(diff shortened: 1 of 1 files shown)


AmeliaBR:
There is an open pull request that is waiting for review about fixing/clarifying the `d` syntax in attributes vs CSS property: https://github.com/w3c/svgwg/pull/374

I suspect your concerns are still valid, but if you're able to, it would be helpful if you compared your fixes against those changes.

SimonSapin:
#374 fixes the same bug that this PR does, in a way that I find satisfactory. It also contains a normative change `path(<string>)` v.s. `<string>`, I don’t really have an opinion about that.

AmeliaBR:
Ok, thanks for checking Simon, and thanks for the effort of the PR even if it turned out to be a duplicate.

PR #623, opened 2019-01-07
fix #609

### LINKED ISSUE #609: text-overflow definition moved from CSS UI to CSS Overflow
text-overflow definition moved from CSS UI to CSS Overflow. We need to update the spec text and references.

svgeesus:


master/text.html
```diff
@@ -5599,8 +5599,8 @@ <h4 id='TextOverflowProperty'>The <span class="property">'text-overflow'</span>
 
   <p class="note">
     See the CSS3 UI specification for the definition of
```
ewilligers: CSS Overflow specification

svgeesus: fixed

svgeesus:
fixed ref to overflow in styling chapter

PR #636, opened 2019-01-21
Add testing widget

svgeesus:
(same as used by CSS WG)

Links spec sections to associated tests, allows creation of test implementation report as needed to exit CR


AmeliaBR:
This doesn't seem to be working...

It's only included in Overview.html, so it doesn't show up at all if you look at the individual chapters.

But even in [single-page view (warning: giant file)](https://svgwg.org/svg2-draft/single-page.html), it doesn't seem to be finding any of the tests.

AmeliaBR:
(Also, we have a class name conflict between our styles and the widget's, both using class name `annotation`.)

Issue #199, opened 2016-07-11
What should happen regarding 1.2 Tiny features that are not mentioned in SVG2 spec?


BigBadaboom:
There are a few references to 1.2 Tiny in the SVG2 spec.  However I have noticed that there are a number of 1.2 Tiny features that AFAICS are not mentioned at all in this spec.  Since 1.2 Tiny reached recommendation status, shouldn't these be mentioned somewhere?  Even if it is only in a deprecated features table?

For example (not exhaustive - just a quick search made)

- `viewport-fill` and `viewport-fill-opacity` properties
- `buffered-rendering` property
- `transform="ref()"`
- `<textArea>`, `<tbreak>`, and related properties
- `editable` on text elements

AmeliaBR:
Tiny 1.2 is mostly being treated as a dead-end fork of SVG.  So consider all of those as deprecated or obsolete.

From what I can tell, most of the references to Tiny in SVG 2 are in the form of annotations about specific features from that spec that we decided to keep.

However, there are a couple back-references to Tiny in the `vector-effects` section and the "Painting vector effects" sections. Probably both need a re-write.  (There's also links to Tiny in the definitions-filters.xml file that we need to update to point to equivalent definitions in SVG 2.  )

Beyond that: If there are specific features from Tiny you want to preserve in SVG 2, that could be considered as a feature request.  From your list, I can say that some are now redundant:

- `viewport-fill` is replaced with CSS `background` properties 
- `buffered-rendering` was dropped in favor of `will-change`
- the multi-line text model has been completely over-hauled so that `<textArea>` is not required

However, a `<tbreak>` element may still prove useful in the future, as could a `editable` option.  

A `ref` type transform, that could reset some or all of the transformation sequence to another SVG, is definitely something I'd be interested in, but it would have to be discussed in the context of the CSS Transforms Level 2 module.

BigBadaboom:
Even if none of those features make it to SVG2. I feel like the spec should at least acknowledge 1.2 Tiny and explain that it is considered a dead fork.  And that SVG2 is a development of 1.1.

At the moment 1.2 is only present as a sort of unacknowledged haunting spirit.  Poor old 1.2 :)

AmeliaBR:
Any suggestions of where to integrate this context? Somewhere in the overview page?

BigBadaboom:
In the Introduction maybe?  Something like:

> 1.4 Relationship to previous versions of this standard
>
[...]

PR #643, opened 2019-02-12
Added paragraph about Tiny 1.2 being deprecated (bug #199)

BigBadaboom:
The suggested text I offered in bug #199 seemed to be acceptable to people,. So I have created a PR based on that. The text here has been slightly modified from the original.

Bug #199 has been marked as deferred to 2.1. But I think it would be a good idea to include this in the first release of SVG 2.

### DIFF
--- a/master/intro.html
+++ b/master/intro.html
@@ -93,6 +93,15 @@ and standards efforts, as described in the following:</p>
   See <a href="access.html">Accessibility Support</a>.</li>
 </ul>
 
+<h2 id="RelationshipToPrevious">Relationship to previous versions of this standard</h2>
+
+This edition of the SVG standard has been developed based on, and built upon, the 1.1 edition released in 2003.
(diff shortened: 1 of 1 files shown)


svgeesus:
@BigBadaboom please [link your GitHub account to your W3C account](https://labs.w3.org/repo-manager/pr/id/w3c/svgwg/643) so that our IPR robot is happy Thanks!

So, what does the WG think of the proposed text? It seems broadly correct to me.

BigBadaboom:
@svgeesus My GitHub account is already linked.  But I don't think it was when I created this PR. Could that be the problem?

AmeliaBR:
@svgeesus, can you follow up within W3C to find out why the IP bot still doesn't like Paul's contribution?  I tried using the "revalidate" option in the [app](https://labs.w3.org/repo-manager/pr/id/w3c/svgwg/643), but it still isn't making the connection.

Issue #649, opened 2019-02-25
Path data: period character should be Unicode U+002E?


danburzo:
In chapter [__9.3. Path data__](https://svgwg.org/svg-next/paths.html#PathDataGeneralInformation):

> The path data syntax is a prefix notation (i.e., commands followed by parameters). The only allowable decimal point is a Unicode U+0046 FULL STOP (".") character (also referred to in Unicode as PERIOD, dot and decimal point) and no other delimiter characters are allowed [UNICODE]. (For example, the following is an invalid numeric value in a path data stream: "13,000.56". Instead, say: "13000.56".)

Should that instead be Unicode U+002E? (46 being the decimal equivalent).

PR #650, opened 2019-02-25
Fix for #649. Incorrect Unicode code point reference

BigBadaboom:
Period/Full-stop is U+002E, not U+0046 (which is a capital F).

For issue #649.


dontcallmedom:

PR #663, opened 2019-04-16
[svg-native] Add SVG Native as an Editor's Draft

litherum:


AmeliaBR:
When we figure out how we're going to handle the build properly, we can get rid of the duplicate .html file. For now, we'll just need to make sure they stay in sync for updates.

litherum:
Thank you!!

PR #682, opened 2019-05-07
desk top => desktop

### LINKED ISSUE #654: Typo in SVG media registration appendix
Hi, I was just reading the MIME registration for SVG and noticed a typo under "Applications that use this media type". There's a "desk top" that probably should be "desktop", as in "...graphical assets in desk top publishing;"

https://www.w3.org/TR/SVG/mimereg.html

ewilligers:
resolves #654

### DIFF
--- a/master/mimereg.html
+++ b/master/mimereg.html
@@ -146,7 +146,7 @@ This media type registration is extracted from Appendix P of the <a href="https:
 Applications that use this media type:</dt>
 <dd>
 <p>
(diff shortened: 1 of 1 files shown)


AmeliaBR:
@svgeesus are there any legal complications of updating the MIME registration. Should we even be republishing it in SVG 2.

(I mean, this is a pretty minor typo fix, but still.)

svgeesus:
If we change the registration appendix then it would need to be resubmitted to IANA and their registration for the Internet Media type would be updated.

I don't see a reason to do that re-submission  for a non-technical change that is arguably not even a typo. 

On the other hand, maybe we will make other changes and resubmit it anyway. If so, I would rather not have to remember to apply this as well. So my inclanation is to 

a) accept the patch
b) not update the IANA registration right now with this very minor and non-technical change
c) if we udate the registration later because of a technical change, this gets rolled in

PR #686, opened 2019-05-07
close #655

### LINKED ISSUE #655: member-svg-media-type@w3.org is no longer active
I found a typo (#654) and tried to email this list as it is written at the bottom of the page i question (https://www.w3.org/TR/SVG/mimereg.html).

But it got bounced with the message that the mailing list is no longer active. So I email site-comments@w3c.org and they directed me here.

I'm not totally sure what makes the most sense, but any email address mentioned in a spec should probably be reachable, even if it redirects to site-comments or similar.

svgeesus:
Set the contact point (which has to be an email address) to the public SVG list; and also point to a GitHub issue which is a better place for any comments.

Oh and some trailing whitespace normalization.

Fixes #655

### DIFF
--- a/master/mimereg.html
+++ b/master/mimereg.html
@@ -159,11 +159,11 @@ Additional information:</dt>
     <dt>File extension(s):</dt>
      <dd>svg
      <p>Note that the extension 'svgz' is used as an alias for 'svg.gz'
-       <cite>[<a href="refs.html#ref-rfc1952">rfc1952</a>]</cite>, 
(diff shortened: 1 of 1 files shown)


svgeesus:
By the way the IANA registraton is required to explain fragment identifiers for this media type, which means we should take care not to break https://www.w3.org/TR/SVG/linking.html#SVGFragmentIdentifiers

AmeliaBR:
Because the reference to fragment types is an internal link in the spec, our build system should warn us if we mess it up.

Issue #692, opened 2019-05-15
mask-type is not listed as a presentational attribute in SVG spec


faceless2:
The "mask-type" attribute is not on the list of presentational attribute in https://svgwg.org/svg2-draft/attindex.html#PresentationAttributes - it is (see https://drafts.fxtf.org/css-masking-1/#the-mask-type)

PR #693, opened 2019-05-21
Add mask-type to list of presentation attributes. Fixes issue #692

dirkschulze:


AmeliaBR:
The presentation attribute should be added to the table in the [styling chapter](https://svgwg.org/svg2-draft/styling.html#PresentationAttributes).

But is it really intended that `mask-type` should be a general presentation attribute? I thought it was an attribute specific to `<mask>` elements.  But then, we have `stop-opacity` and `flood-color`, and all sorts of other presentation attributes that only apply to one element but are valid everywhere, so I guess that's OK.  That table only defines restrictions when there is a conflict with other attributes.

dirkschulze:
@AmeliaBR ping

PR #694, opened 2019-05-21
Add transform-origin to list of presentation attributes. Fixes #688

### LINKED ISSUE #688: Support transform-origin as a presentation attribute
Based on a [question from a colleague](https://mobile.twitter.com/Mamboleoo/status/1126040958045491200), I discovered that 3 of 4 major browsers (Chromium, WebKit, and EdgeHTML) currently support `transform-origin` as a presentation attribute. EdgeHTML requires units; Chromium and WebKit support it with unitless numbers, consistent [with the usual SVG 2 rules for presentation attributes](https://svgwg.org/svg2-draft/types.html#syntax). Firefox does not support `transform-origin` as a presentation attribute.

Based on the current support, I'd recommend adding a presentation attribute for `transform-origin` to SVG 2.

dirkschulze:


AmeliaBR:
Same comment as for #693, need an entry in https://svgwg.org/svg2-draft/styling.html#PresentationAttributes (which isn't generated automatically because it needs to exclude certain values).

dirkschulze:
@dirkschulze ping

Issue #335, opened 2017-07-21
SVG2 path data coordinates are currently spec'd as integers.


fuchsia:
I'm sure you're aware, but just in case you're not: the current version of [Path Data Grammar in SVG 2](https://svgwg.org/svg2-draft/paths.html#PathDataBNF) specifies coordinates as signed integers. Support for decimals and exponents seems to have been lost in the haze of rewrites.

AmeliaBR:
It has been noted, but I don't think we had a dedicated issue.

See also:

- https://github.com/w3c/svgwg/issues/286
- https://github.com/w3c/svgwg/issues/331
- https://lists.w3.org/Archives/Public/www-svg/2017Apr/0020.html
- https://github.com/w3c/svgwg/issues/325

fuchsia:
If it helps, I think the SVG 1.1 grammar can be written as:
```
sign       ::= "+" | "-"
digits     ::= [0-9]+
unsigned   ::= digits
integer    ::= sign? unsigned
exponent   ::= [Ee] integer
fraction   ::= ( digits ( "." digits? )? ) | ( "." digits )
float      ::= sign? fraction exponent?
coordinate ::= float
```
But there are many other ways to do it. 

IMPORTANT, the first two arguments for `eliptical_arc_argument`/`elliptical_arc_closing_argument` should be `unsigned` and the third `integer`:
```
elliptical_arc_argument::= 
    unsigned comma_wsp? unsigned comma_wsp? integer comma_wsp 
    flag comma_wsp? flag comma_wsp? coordinate_pair

elliptical_arc_closing_argument::= 
    unsigned comma_wsp? unsigned comma_wsp? integer comma_wsp 
    flag comma_wsp? flag comma_wsp? closepath
```
(Again, I'm going by the 1.1 grammar.) 

Also, you need to decide whether the argument to `bearing_argument_sequence` is `unsigned` or `integer`. (I've deliberately not used `number` as a symbol so we can spot similar errors.)

The presentation of `fraction` is slightly different to that in SVG 1.1, but matches the one I use for character-by-character parsers and as a regex (`/[-+]?(?:\d+(?:\.\d*)?|\.\d+)(?:[Ee][+-]?\d+)?/`).

The above syntax makes floats like `4.` legal. That matches javascript but differs from CSS, where I think `4.` is an invalid number.

AmeliaBR:
Thanks for that @fuchsia .

[...]

PR #697, opened 2019-05-22
Align path number parsing with CSS Numbers. Fixes #331

### LINKED ISSUE #331: Floating point number format discordance (SVG vs CSS vs browsers)
Summary:
  - SVG path grammar defines a number text format not found in CSS
  - some browsers only support the more restrictive CSS number formats
  - a sample set of SVG files did find uses of all the SVG formats
  - but the 'missing' format may not be used widely
  - what do with this discordance?
  - (added images for Edge/Chrome/FireFox)
  - (added codepen link for reviewing problem)

Having familiarized myself with the SVG grammar for floating point numbers I was surprised when a browser balked at drawing a manually-typed path that included a coordinate number "-41."

The SVG specs from 2011 and 2001 permit a full set of text representations for floating point numbers: "0", "0.0", ".0" and "0."  In comparison, CSS permits only three formats, "0", "0.0", ".0", explicitly omitting 
the format "0.". 

From testing SVG path strings in browsers it would seem that both FireFox and Chrome have opted to implement only the CSS definition of floating point numbers.  They do not permit numbers of the form "0." as required by SVG.  Microsoft's Edge, the more recent implementation of SVG, *does conform* to the SVG spec and implements all four formats.  (the status of Safari is unknown to me)

.
Now the format "nn." could be considered a 'degenerate' or redundant form, having the obvious equivalent representations "nn.0" and "nn".  If the smallest set of formats was desired, then ".0" is replaceable by "0.0", resulting in only two formats needed, "nn" and "n.n"  But if providing representations that minimize number text lengths is desired, then ".0" is justified whereas "0." is not.  Perhaps this latter reasoning - text length - was used in the CSS specification for deciding which formats to include?

.
A hopelessly inadequate sample set of 22K SVG files was examined to see which floating point number formats were in use in the wild.  This file set *did* contain numbers of the form "nnn."  However, and 
significantly, these were found in only one file which is *internal* to the Inkscape application.  These numbers are never presented to a web browser.

Again, this sample set is not a valid cross-section of SVG files.  However, it is _possible_ that SVG files in the wild really have  restricted themselves to the smaller set of number formats supported across web browsers.  That the web has likely once again worked-around incomplete implementations is not surprising.

.
Some browsers have not implemented the now 16 year-old SVG specification.  Similarly amazing is that the SVG test cases, and currently the  W3C Web Platform Tests, never tested something as fundamental as the specified SVG number formats.

.
**The question of what to do now remains.**  Should SVG insist on application support for a simple number format, which requirement has been published for 16 years?   Or should the 'guideline' be invoked that specifications should drop those things not implemented by more than one application in a compatible manner?

.
Images from [Codepen demo](https://codepen.io/tshinnic/full/XgOvVP/): &nbsp; (Note: red is 'bad')
Edge svg_svg2_fp_number_grammar_issue_edge Chrome svg_svg2_fp_number_grammar_issue_chrome FireFox svg_svg2_fp_number_grammar_issue_firefox

.
Notes:

1) SVG number format definition:
    https://www.w3.org/TR/SVG/paths.html#PathDataBNF
            (then search for "floating-point-constant")

2) CSS floating point formats definition:
  https://www.w3.org/TR/css3-values/#numbers
  https://drafts.csswg.org/css-syntax-3/#typedef-number-token
  https://drafts.csswg.org/css-syntax-3/#number-token-diagram
>      When written literally, a number is either an integer,
>      or zero or more decimal digits followed by a dot (.)
>      followed by one or more decimal digits and optionally
>      an exponent composed of "e" or "E" and an integer.
(and this does not allow "123." )

3) Codepen test for number format support
      https://codepen.io/tshinnic/full/XgOvVP/

dirkschulze:


AmeliaBR:
This looks like the fix to #335, not #331.  Otherwise, looks good to me, but I'm not sure how to amend the PR so the correct issue gets fixed when it is merged.

dirkschulze:
@AmeliaBR It is both actually. #331 is related to paths as well. (All other occurrences reference CSS Number already.)

AmeliaBR:
Ok. Our resolution [from last June](https://github.com/w3c/svgwg/issues/331#issuecomment-396359891) was:

> Use CSS numbers, allow UAs to implement SVG 1.1 numbers but strongly discourage to use them

So I guess this takes care of the first half of that, by making parsing match CSS.  Do you want to add a paragraph regarding the second part?  Possible wording:

> The SVG 1 path grammar was defined in a way that allowed a trailing decimal point without any decimal digits (like `23.` ).  SVG 2 harmonizes number parsing with CSS, and makes this invalid.  However, user agents may continue to accept numbers with trailing decimal points when parsing is unambiguous. Authors and authoring tools must not use this number format.

dirkschulze:
@AmeliaBR Updated PR

Issue #110, opened 2016-04-20
Make "show" widget in element definitions boxes play nicely with screen readers


AmeliaBR:
In SVG2 spec element boxes, the "show" to expand categories is being read twice by JAWS/Firefox (and maybe others).

dstorey:
I haven't tested it yet, but it may be because the "show" is included in generated content, which is read by screen readers. I don't see where it is included the second time though as there is no text node.

This would probably be better as a details/summary with polyfill as there is no aria in the current markup to show it is a complex control with expanded state, and there is no way to collapse it again after expanding it.

PR #714, opened 2019-07-14
Make "show" widget in element definitions boxes play nicely with screen readers

takenspc:
This PR

- Makes show widget keyboard accessible
- Makes arrow icon decorative

closes https://github.com/w3c/svgwg/issues/110

### DIFF
--- a/master/style/expanders.js
+++ b/master/style/expanders.js
@@ -1,31 +1,30 @@
-function expand(e) {
-  e = e.parentNode;
-  e.parentNode.removeAttribute('class');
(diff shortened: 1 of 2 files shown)


AmeliaBR:

AmeliaBR:
Thank you! This looks great.

I'm not sure how you came upon that old To-Do bug, but glad you took the time to do something about it!

AmeliaBR:
Well, it looked great on my local build, but on the live version we're still getting the old CSS. Which means it's showing up as "showshow >>". 😞

PR #729, opened 2019-08-28
Link to definition of color for stop-color syntax

ewilligers:
We previously defined the stop-color syntax as
`currentColor | <color> <icccolor>`
which would have suggested that `black icc-color(name)` is
a valid stop-color but `black` is not.

### DIFF
--- a/master/propidx.html
+++ b/master/propidx.html
@@ -210,9 +210,7 @@
         </tr>
         <tr>
           <th><a>'stop-color'</a></th>
-          <td>currentColor |<br />
-           <a>&lt;color&gt;</a>
-           [<a>&lt;icccolor&gt;</a>]</td>
+          <td>&lt;‘<a>'color'</a>’&gt;</td>
           <td>black</td>
           <td><a>'stop'</a> elements</td>
           <td>no</td>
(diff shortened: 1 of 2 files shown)


ewilligers:
Tests:
https://wpt.fyi/results/svg/pservers/parsing

AmeliaBR:
Woah, that's been wrong for a while.

A few additional changes required, to propidx.html (which is sadly still not auto-generated), and maybe to the introduction text about `stop-color` which talks about `currentColor` and ICC colors.

ewilligers:
> A few additional changes required, to propidx.html (which is sadly still not auto-generated), and maybe to the introduction text about `stop-color` which talks about `currentColor` and ICC colors.

Done

master/propidx.html
```diff
@@ -210,9 +210,7 @@ <h1>Property Index</h1>
         </tr>
         <tr>
           <th><a>'stop-color'</a></th>
-          <td>currentColor |<br />
-           <a>&lt;color&gt;</a>
-           [<a>&lt;icccolor&gt;</a>]</td>
+          <td>&lt;‘<a>'color'</a>’&gt;</td>
```
dirkschulze: I don't see a mention that we are dropping `icc()` in the SVG2 change log. Do you know if we have a resolution for that? We probably have.

ewilligers: I don't know if there is a resolution. This edit simply makes stop-color consistent with other color properties.

The spec is currently inconsistent as it suggests `#A8F` and `white` are valid stop-color values, but the grammar requires something like`<color> icc-color(<name> [ ',' <number> ]+ )`.

PR #740, opened 2019-09-23
Remove onfocusin and onfocusout event attributes.

### LINKED ISSUE #715: Defer onfocusin and onfocusout
In SVG 2 spec, elements have graphical event attributes (onfocusin and onfocusout) while SVG 2 spec doesn't have formal definition of them.

Example:
https://svgwg.org/svg2-draft/struct.html#SVGElement
> graphical event attributes — ‘onfocusin’, ‘onfocusout’

On the other hand, HTML Standard has a few issues about formalization of onfocusin and onfocusout.

- https://github.com/whatwg/html/issues/3514
- https://github.com/whatwg/html/issues/4542

I think it is better to drop onfocusin and onfocusout for a while and update global event attributes once onfocusin and onfocusout were defined in HTML Standard.

heycam:
Fixes #715

### DIFF
--- a/master/changes.html
+++ b/master/changes.html
@@ -973,6 +973,7 @@ have been made.</p>
   <ul>
     <li>Remove requirements on <a>'pointer-events'</a> for raster images to check individual pixels for transparency.</li>
     <li>Remove <code>zoomAndPan</code> attribute and related text.</li>
+    <li>Removed the onfocusin and onfocusout event attributes.</li>
   </ul>
 </div>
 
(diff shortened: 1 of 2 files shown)


dirkschulze:
@heycam The resolution seems to imply that we want to remove them when we can reference them from HMTL instead. IIRC this patch just removes the focus attributes. Did you check if they were added to HTML?

heycam:
@dirkschulze I didn't check, but that haven't been yet. I read the resolution as asking for the attributes to be removed, but they will come back automatically at some point when they're defined in HTML. Do you suggest blocking on HTML before merging this?

dirkschulze:
@heycam Reading the IRC log it seems we resolved to remove even if they aren't in HMTL yet.

PR #743, opened 2019-10-01
Delete Makefile

bernadeth10:


dontcallmedom:

dirkschulze:
Why would we need to remove the make file? Looking at it, it should still work and create a versioned spec.

bernadeth10:
BBFdelete everything
we dont make nothing pls delete 

Em Sex 19/10/04 14:18, Dirk Schulze notifications@github.com escreveu:

Why would we need to remove the make file? Looking at it, it should still work and create a versioned spec.

AmeliaBR:
That's the main build file for the repo. It may be old-fashioned, but it still works.

PR #748, opened 2019-11-09
Rebase the SVG Native spec on top of SVG 2

litherum:
This reorganizes the SVG Native spec to match SVG 2. I don't believe there are any (meaningful) behavior changes.

While doing this, I also included all the resolutions the SVG WG has made. A link to each resolution is included in a comment in the source.

This should catch the SVG Native spec up with current discussion.

### DIFF
--- a/specs/svg-native/index.bs
+++ b/specs/svg-native/index.bs
@@ -54,6 +54,8 @@ There are numerous differences between the SVG Native format and the [[!SVG2]] f
 
 Note: If a document uses parts of SVG that aren't part of SVG Native, then it isn't a conformant SVG Native document. When an SVG Native renderer opens such a file, it should ignore all of those parts. Therefore, all SVG Native renderers should produce the same rendering for the same document. However, there may be renderers which intentionally decide to honor parts of SVG that aren't part of SVG Native. In order to make it more likely that a given document's rendering is predictable, even in these renderers, SVG Native authoring tools are recommended to always emit documents that conform to this specification. The SVG Working Group is considering producing a validator to determine whether a given document is conformant to the SVG Native format.
 
+ISSUE: The file type and mime type is being discussed in <a href="https://github.com/w3c/svgwg/issues/664">this GitHub issue</a>.
+
 SVG Native is a standalone file type, expected to be rendered with a dedicated-purpose renderer. Therefore, SVG Native content must not be present as part of a larger XML (or HTML) document. If it is present as part of a larger XML (or HTML) document, the content should be interpreted as SVG proper.
 
 Note: This means that browsers encounting a SVG Native root element within the DOM must not interpret it as SVG Native, even if it has the 'baseAttribute' attribute set to <code>native</code>. If a Web author desires to use SVG Native on the Web, the <{img}> element may be used instead. This matches other native image formats such as [[!PNG]]. A Web browser should implement SVG Native by linking with the system's SVG Native facilities. This matches the implementation of other native image formats.
@@ -66,7 +68,7 @@ User agents may limit the reference depth of references to implementation-depend
 
 XSL Processing is forbidden.
 
-XML Entities and CDATA sections may be present.
+A valid SVG Native document must not contain any of the external or internal XML DTD subset.
 
 ISSUE: Is it worth mentioning the 'HTML dialect' of SVG?
 
--- a/specs/svg-native/index.bs
+++ b/specs/svg-native/index.bs
@@ -120,10 +120,21 @@ Percentage length values are not supported by SVG Native.
 <a lt="relative length">Relative units</a> are not supported by SVG Native.
 <!-- https://www.w3.org/2019/07/01-svg-minutes.html -->
 
+Geometry Properties {#geometryproperties}
+=========================================
+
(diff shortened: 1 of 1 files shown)


specs/svg-native/index.bs
```diff
@@ -64,62 +70,118 @@ The file extension for SVG Native is <code>.svn</code>. The UTI for SVG Native i
...
+
+<a lt="relative length">Relative units</a> are not supported by SVG Native.
+<!-- https://www.w3.org/2019/07/01-svg-minutes.html -->
+
+Geometry Properties {#geometryproperties}
+=========================================
+
+The grammar for the 'cx', 'cy', 'r', 'rx', 'ry', 'x', 'y', 'width', and 'height' properties is:
+
+<code><<number>> | <<length>> | <<percentage>></code>
```
AmeliaBR: What is the purpose of re-writing the grammar here?

Percentages are already forbidden above.

The allowance of "number" comes out of the SVG rules on parsing presentation attributes, although it might be useful to add a clarifying note about it. (But note that the parsing rule also affects other presentation attributes, e.g., stroke-width.) It's interpretation as a px value is likewise standard SVG.

I think it would be more consistent with the rest of the spec to forbid "keyword values" for these properties rather than writing a new grammar.

specs/svg-native/index.bs
```diff
@@ -128,128 +190,88 @@ Only the following units are supported:
 - ''cm''
 - ''in''
 
+The 'transform' property is only supported by SVG Native on the following elements:
```
AmeliaBR: What is the purpose of this exception? What are you excluding & why?

specs/svg-native/index.bs
```diff
@@ -128,128 +190,88 @@ Only the following units are supported:
...
-
-The ''scroll'' value on the 'overflow' property behaves identically to 'hidden'.
-
-The 'clip-path' attribute must not be present on a <{clipPath}> element.
-
-Every <{clipPath}> element must have either exactly zero or exactly one child.
-
-The ''scroll'' and the ''auto'' values of the 'overflow' property are forbidden.
-
-Filter Effects {#filters}
```
AmeliaBR: I know that filters aren't defined in SVG anymore, but the `filter` property is listed as a required CSS feature, so you will need to still explicitly exclude it.

And, since the filter elements are still SVG namespace elements, and you have explicitly talked about clipPath and mask elements, maybe want to keep something about no filter support.

(Unless I missed a decision where we decided to *allow* filters…?)

specs/svg-native/index.bs
```diff
@@ -64,62 +70,118 @@ The file extension for SVG Native is <code>.svn</code>. The UTI for SVG Native i
...
+- 'clip'
+
+The ''scroll'' value on the 'overflow' property behaves identically to 'hidden'.
 
-Basic Data Types and Interfaces {#types}
-========================================
+The 'clip-path' attribute is not supported on a <{clipPath}> element.
 
-Chapter is unchanged.
+Every <{clipPath}> element must have either exactly zero or exactly one child. Anything more than that is not supported by SVG Native.
```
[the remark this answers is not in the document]
dirkschulze: I think we did discuss it on an GitHub issue at least. Need to check if we have a resolution for it though. @AmeliaBR  IMO it is reasonable for the editor to do changes like this. The spec is not in a CR state yet. But of course you could bring it up in one of the next WG meetings.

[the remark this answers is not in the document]
litherum: It's from line 192 in the original document. It came from the original draft spec. I don't think we discussed it in the WG. If you would like to, please open an issue and we can add it to the agenda for an upcoming call.

specs/svg-native/index.bs
```diff
@@ -64,62 +70,118 @@ The file extension for SVG Native is <code>.svn</code>. The UTI for SVG Native i
...
+<!-- https://www.w3.org/2019/07/15-svg-minutes.html -->
+
+Conditional processesing attributes and elements are not supported in SVG Native. These include the <{switch}> element.
+
+<!-- There are too many things which are unobservable in SVG Native. Don't need to list them all out. -->
 
-Section 5.8 "Conditional Processing" is deleted.
+The <{meta}> and <{metadata}> elements are not supported in SVG Native.
+
+Elements in foreign namespaces, except the xlink (<code>http://www.w3.org/1999/xlink</code>) namespace, are not supported in SVG Native.
```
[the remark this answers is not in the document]
litherum: I don't think `<meta>` should be supported. It should be ignored, just like all the other stuff that isn't supported.

specs/svg-native/index.bs
litherum: From what I could tell, SVG 2 allows the transform attribute on any element. However, SVG 1.1 doesn't allow that. Allowing it anywhere makes implementation more difficult and confusing; this restriction simplifies it.

specs/svg-native/index.bs
litherum: I believe I excluded it in the "SVG Modules" modules section above.

> No other SVG modules are supported by SVG Native. This includes:
> - <a href="https://www.w3.org/TR/filter-effects-1/">CSS Filter Effects</a>

Issue #752, opened 2019-11-19
Subtleties of the path data grammar


dalboris:
As a lot of us know, the path data grammar is a little convoluted for the sake of path data conciseness.

For example, the spec makes it clear that something like "M0.6.5L100-200" is an intentionally valid path data, equivalent to "M 0.6 0.5 L 100 -200".

In this GitHub issue, I'd like to dig deeper in the rabbit hole, by giving more examples of contrived (or not so contrived) examples, and asking the following three questions for each:

1. Is it currently formally allowed by the grammar?
2. Is it intentionally allowed (or disallowed)?
3. If answers to 2. is "no", should we change the grammar to reflect intent?

I'll just start with a few examples about whitespaces, and we can add more later. I name the examples so we can reference them. This GitHub issue could be used as clarification of intent.

References:

SVG 1.1 grammar: https://www.w3.org/TR/SVG11/paths.html#PathDataBNF
SVG 2-draft grammar: https://svgwg.org/svg2-draft/paths.html#PathDataBNF

# Examples about whitespaces (W)

## Example W1 - trailing whitespaces after path

```
"M1 2L3 4 "
         ^
```

**1. Is it currently formally allowed by the grammar?**

I believe it is allowed by SVG 1.1 grammar, but forbidden by the SVG 2 draft. Relevant parts of the grammar:

```
SVG 1.1      svg-path ::= wsp* moveto-drawto-command-groups? wsp*
SVG 2-draft  svg_path ::= wsp* moveto? (moveto drawto_command*)?
```

(of course, one has to dig deeper into SVG 2-draft's definition to conclude)

**2. Is it intentionally disallowed in SVG 2?**

I hope not. Whitespaces can be used if we prefer to place the closing `"` on its own line, for example, for manually edited SVGs. Besides, leading whitespaces are allowed, so this would be inconsistent.

**3. Should we change the grammar to reflect intent?**

I personally think we should.

## Example W2 - trailing whitespaces after command

```
"M1 2 L3 4"
     ^
```

**1. Is it currently formally allowed by the grammar?**

I believe it is allowed by SVG 1.1 grammar, but forbidden by the SVG 2 draft. Relevant parts of the grammar:

```
SVG 1.1      moveto-drawto-command-group ::= moveto wsp* drawto-commands?
SVG 2-draft  svg_path ::= wsp* moveto? (moveto drawto_command*)?
```

**2. Is it intentionally disallowed in SVG 2?**

[...]

Issue #751, opened 2019-11-18
Use shorter symbol names in SVG path data grammar


dalboris:
# TL;DR

I propose the following grammar. It is exactly the same as the current one, but with shorter symbol names to make rules fit in single lines, increasing readability.

```
path ::= wsp* moveto? (moveto command*)?

command ::=
    moveto
    | closepath
    | lineto
    | hlineto
    | vlineto
    | ccurveto
    | scurveto
    | qcurveto
    | tcurveto
    | arcto

moveto    ::= ("M" | "m") wsp* coord_2s
closepath ::= ("Z" | "z")
lineto    ::= ("L" | "l") wsp* coord_2s
hlineto   ::= ("H" | "h") wsp* coords
vlineto   ::= ("V" | "v") wsp* coords
ccurveto  ::= ("C" | "c") wsp* coord_6s
scurveto  ::= ("S" | "s") wsp* coord_4s
qcurveto  ::= ("Q" | "q") wsp* coord_4s
tcurveto  ::= ("T" | "t") wsp* coord_2s
arcto     ::= ("A" | "a") wsp* arcto_args

coord     ::= sign? unsigned
coord_2   ::= coord cw? coord
coord_4   ::= coord cw? coord cw? coord cw? coord
coord_6   ::= coord cw? coord cw? coord cw? coord cw? coord cw? coord
arcto_arg ::= unsigned cw? unsigned cw? coord cw flag cw? flag cw? coord_2

coords     ::= coord     | (coord     cw? coords)
coord_2s   ::= coord_2   | (coord_2   cw? coord_2s)
coord_4s   ::= coord_4   | (coord_4   cw? coord_4s)
coord_6s   ::= coord_6   | (coord_6   cw? coord_6s)
arcto_args ::= arcto_arg | (arcto_arg cw? arcto_args)

unsigned ::= frac exp?
frac     ::= (digit* "." digit+) | digit+
exp      ::= ("e" | "E") sign? digit+

cw    ::= (wsp+ ","? wsp*) | ("," wsp*)

sign  ::= "+" | "-"
flag  ::= "0" | "1"
wsp   ::= #x9 | #xA | #xC | #xD | #x20
digit ::= "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9"
```

# Long Version

I would like to suggest purely stylistic changes to the path data grammar to make it more readable. Importantly, I believe the current symbol names are way too long, and that this is the most significant barrier to readability.

[...]

PR #778, opened 2020-03-02
Fix missing trailing wsp after path commands

JeremiePat:
Hello SVG WG.

I'm currently implementing an SVG Path parser and it appears that the SVG2 path grammar is a little inconsistent with SVG 1.1 and with other current implementations (AFAIK those from web browsers).

Many details have been discussed in issue #752 (mostly point 1 and 2 raised by @dalboris)

This PR is a proposed fix for the missing trailing white spaces after commands and at end of path. The intent is to prevent a breaking change to the SVG Path grammar. If the SVG WG think this breaking change should be kept, so it should be clearly stated (exactly like the warn done for numbers with trailing dot). If so, I would be glad to update this PR accordingly.

+@dalboris

### DIFF
--- a/master/paths.html
+++ b/master/paths.html
@@ -800,7 +800,11 @@ SVG arc notation to a format that uses center points and arc sweeps.
 
 <p>SVG path data matches the following EBNF grammar.</p>
 <pre class='grammar ready-for-wider-review'>
-svg_path::= wsp* moveto? (moveto drawto_command*)?
+svg_path::= wsp* (moveto wsp* drawto_command_group*)? wsp*
+
+drawto_command_group::=
+    drawto_command
+    | drawto_command wsp* drawto_command_group
 
 drawto_command::=
     moveto
--- a/master/paths.html
+++ b/master/paths.html
@@ -800,11 +800,7 @@ SVG arc notation to a format that uses center points and arc sweeps.
 
 <p>SVG path data matches the following EBNF grammar.</p>
 <pre class='grammar ready-for-wider-review'>
-svg_path::= wsp* (moveto wsp* drawto_command_group*)? wsp*
-
-drawto_command_group::=
-    drawto_command
-    | drawto_command wsp* drawto_command_group
+svg_path::= wsp* (moveto (wsp* drawto_command)*)? wsp*
 
 drawto_command::=
     moveto


JeremiePat:
Ah! This my first PR to W3C so I forgot to link my GitHub account. It is done, feel free to revalidate the check 😉

dalboris:
Good initiative. I was waiting for more feedback from the committee to know whether I should submit a PR to fix/improve the grammar.

In general, I personally favor a concise grammar, more info here: 

https://github.com/w3c/svgwg/issues/751

In the specific case of the trailing whitespaces, here is what I advocate:

https://github.com/w3c/svgwg/issues/751#issuecomment-555163624

```
svg_path ::= wsp* (moveto (wsp* drawto_command)*)? wsp*
```

It is equivalent to your:

```
svg_path::= wsp* (moveto wsp* drawto_command_group*)? wsp*
drawto_command_group::=
    drawto_command
    | drawto_command wsp* drawto_command_group
```

But without the additional symbol `drawto_command_group` which makes the grammar three lines longer and harder for humans to parse. Less symbols = less cognitive load, in my opinion.

JeremiePat:
I totally agree with you that a shorter grammar definition would be helpful to spec implementer. It took me three iterations banging my head on a wall to start having a reasonable understanding of the path grammar. I would love to have seen issue #751 earlier to avoid that 😋. FWIW I would support a shorter, cleaner, version of the grammar either in the next iteration of the SVG Spec or within the dedicated [Path Spec](https://svgwg.org/specs/paths/).

For that fix, the first suggestion I made was a direct inspiration from SVG 1.1, but I really like your suggestion to shorten this PR. However, I'm not super at ease with nesting parenthesis. This is one of this case that make things a little harder to process for a human brain (at least, mine)

I was considering something like this (yes, still an additional symbol):

```
svg_path::= wsp* (moveto wsp* drawto_command_group*)?
drawto_command_group::= (drawto_command wsp*)+
```

There is also the option to add a `wsp*` symbol at the end of each command definition and simplify the grammar to:

```
svg_path::= wsp* (moveto drawto_command*)?
```

dalboris:
> It took me three iterations banging my head on a wall to start having a reasonable understanding of the path grammar. 

Same here :)

Thx for the amend to your PR! I do find this 1-liner version the most readable despite the nested parentheses, but it's obviously very subjective.

Note that in your 2-line suggestion above, the `+` in `(drawto_command wsp*)+` is redundant with the `*` in `drawto_command_group*` (in other words, `(A+)*` and `A*` are equivalent), so you could simplify the suggestion to:

```
svg_path::= wsp* (moveto wsp* drawto_command_group*)?
drawto_command_group::= drawto_command wsp*
```

(in which case the name `drawto_command_group` should obviously be changed, e.g. to `command_wsp`, but then since it's so simple and only used once... well, we might as well inline it and avoid finding a good name ;) )

dalboris:
Re adding `wsp*` at the end of each command definition: we could even go further and add `wsp*` also at the beginning of each command. This way, we would have:

```
svg_path::= (moveto drawto_command*)?
```

This restores symmetry, otherwise readers might otherwise wonder which there is `wsp*` at the end of each command, but not at the beginning.

Either way, I'm not sure it's worth the added noise in each command. Reading such grammar, I would probably tell myself: why on earth didn't they factorize these `wsp*`?

dalboris:
@tabatkins @caribouW3 Would one of you, or another member of the committee with merge power, be willing to review and merge this change? It's a one-line fix of a bug in the grammar, that was unintentionally introduced when modifying the grammar from SVG 1.1 to 2.0. 

This issue has already been reported several times, the last one being: https://github.com/w3c/svgwg/issues/929

dalboris:
Thanks @caribouW3 for merging!

Note: it is also now possible to close this issue: https://github.com/w3c/svgwg/issues/807

PR #796, opened 2020-04-30
textPath example: Replace `path` attribute with `d`

speleo3:
No name change for `d` attribute. See ANNOTATION 1:
https://www.w3.org/TR/SVG/paths.html#PathElement

### DIFF
--- a/master/text.html
+++ b/master/text.html
@@ -4413,7 +4413,7 @@
 style="font-weight:bold; color:red">transform="translate(25,25)"</span>&gt;
     &lt;defs&gt;
       &lt;path id="path1" <span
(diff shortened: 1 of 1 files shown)


longsonr:
This is wrong, the attribute is path. https://www.w3.org/TR/SVG/text.html#TextPathElementPathAttribute note that Firefox has implemented this.

speleo3:
@longsonr Your link points to the attribute for the `textPath` element, not the `path` element.

textPath-vs-path

This PR fixes the attribute on the `path` element.

example-path-path

longsonr:
Ahh, it's the path examples, yes I agree then.

PR #802, opened 2020-07-01
Drop the [NoInterfaceObject] for SVGPathSegment

foolip:
This has been renamed in Web IDL:
https://heycam.github.io/webidl/#LegacyNoInterfaceObject

However, if the interface is implemented, it would be best to expose it for feature detection.

### DIFF
--- a/specs/paths/master/Overview.html
+++ b/specs/paths/master/Overview.html
@@ -1368,7 +1368,7 @@ commands contribute to path length calculations.</p>
 The <a>SVGPathSegment</a> interface is a base interface that corresponds to a
 single command within a path data specification.
 
-<pre class="idl">[NoInterfaceObject]
+<pre class="idl">
 interface <b>SVGPathSegment</b> {
   DOMString type;
   sequence&lt;float> values;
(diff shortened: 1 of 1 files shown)


foolip:
@tabatkins do you know who I can poke for review of a change like this?

specs/paths/master/Overview.html
```diff
@@ -1368,10 +1368,10 @@ <h3 id="InterfaceSVGPathSegment">Interface SVGPathSegment</h3>
 The <a>SVGPathSegment</a> interface is a base interface that corresponds to a
 single command within a path data specification.
 
-<pre class="idl">[NoInterfaceObject]
+<pre class="idl">
 interface <b>SVGPathSegment</b> {
```
saschanaz, line 1372: Looking at how this is used, maybe this can even be a dictionary.

foolip: I think you're right, that would be fine in this case, and would avoid the need to make `values` a `FrozenArray<float>` as one would need to do to keep this an interface.

saschanaz:
SVGPathElement is now defined by SVG 2, maybe this spec should probably be dropped.

PR #806, opened 2020-08-18
[svg-paths] Fix IDL syntax errors

saschanaz:


dontcallmedom (approved):

saschanaz:
SVGPathElement is now defined by SVG 2, maybe this spec should be dropped.

saschanaz:
Per #818 there's no intention to maintain it, closing.

PR #816, opened 2020-11-22
Number grammar adjusted

### LINKED ISSUE #815: [Paths/Grammar] Number grammar does not include point for floating-point integers
This issue concerns a wrong grammar definition in `specs/paths/master/Overview.html` in the section on *The grammar for path data*. The link to which can be found [here](https://www.w3.org/TR/SVG/paths.html#PathDataBNF).

One will find the grammar definition for a *number* given as:
```
number ::= ([0-9])+
```
This is a fault because it does not include floating-point integers.

Probably this should be:
```
number ::= ( ([0-9])+ | ([0-9])+ "." | ([0-9])+ "." ([0-9])+ )
```

coastalwhite:
Number grammar now includes floating-point integers instead of only integers.

Fixes #815 
> This issue concerns a wrong grammar definition in `specs/paths/master/Overview.html` in the section on _The grammar for path data_. The link to which can be found [here](https://www.w3.org/TR/SVG/paths.html#PathDataBNF).
> 
> One will find the grammar definition for a _number_ given as:
> 
> ```
> number ::= ([0-9])+
> ```
> 
> This is a fault because it does not include floating-point integers.
> 
> Probably this should be:
> 
> ```
> number ::= ( ([0-9])+ | ([0-9])+ "." | ([0-9])+ "." ([0-9])+ )
> ```

### DIFF
--- a/specs/paths/master/Overview.html
+++ b/specs/paths/master/Overview.html
@@ -1202,7 +1202,7 @@ coordinate_pair::= coordinate comma_wsp? coordinate
 coordinate::= sign? number
 
 sign::= "+"|"-"
-number ::= ([0-9])+
+number ::= ( ([0-9])+ | ([0-9])+ "." | ([0-9])+ "." ([0-9])+ )
 flag::=("0"|"1")
 comma_wsp::=(wsp+ ","? wsp*) | ("," wsp*)
 wsp ::= (#x9 | #x20 | #xA | #xC | #xD)


SebastianZ:
I think that could also be written like this, removing the redundancies and making it a bit more condensed:

```
number ::= [0-9]+ ("." [0-9]*)?
```

Having said that, the [specification currently claims](https://svgwg.org/svg2-draft/paths.html#PathDataBNF) to use [EBNF](https://www.cl.cam.ac.uk/~mgk25/iso-14977.pdf) for the syntax definition. Though, as far as I can see, neither the syntax currently present in the spec., nor the syntaxes defined here are actually EBNF. They are rather a mixture of different BNF syntaxes and [CSS Values](https://drafts.csswg.org/css-values-4/) or regular expressions syntax.
Though the [references link to ABNF](https://svgwg.org/svg2-draft/refs.html#NormativeReferences).

Would be nice if someone clarified what syntax should actually be used.

Sebastian

coastalwhite:
I totally agree on the point that the type of syntax is not very clear. This should be clarified, maybe I will look into this.

To get further into this specific problem. It is actually way more complicating than I originally thought.

> ```
> number ::= [0-9]+ ("." [0-9]*)?
> ```
This grammar and the one I suggested would not work, since it needs to support basically 3 different formats also with an possible extension. Furthermore, the after the point there needs to be one or more digits in your grammar, otherwise we wouldn't be able to (when adding the other formats) distinguish between parsing "5.2" to _5.0_ and _0.2_ or parsing it _5.2_.

The three different formats are:
- Plain integers _1_, _2_, _42_, etc.
- Plain floating point numbers _1.0_, _42.24_, _123.456_, etc.
- Unopend floating point numbers _.5_, _.1_, _.42_, etc. This is also explained in this piece of text from just below the syntax:
> Similarly, for the string "M 0.6.5", the first coordinate of the "moveto" consumes the characters "0.6" and stops upon encountering the second decimal point because the production of a "coordinate" only allows one decimal point. The result is that the first coordinate will be "0.6" and the second coordinate will be ".5".

The extension which I forgot is found in `0.12e10`. I don't know how the concensus is on this extension, but it is supported in most major browsers.

A syntax which would take all of this into account would then look like:
```
number ::= ( ([0-9]+ ("." [0-9]+)?) | ("." [0-9]+) ) exponent_extension?
exponent_extension ::= ("e" | "E") [0-9]+
```

Or alternatively you could write it like this:
```
number ::= ( ([0-9]+ float_ext?) | float_ext ) exponent_ext?
float_ext ::= "." [0-9]+
exponent_ext ::= ("e" | "E") [0-9]+
```
Personally I think this makes it way more readable.

This syntax leads to supporting all different formats and allows for unambiguous interpretation within the standard when parsing.

As said please don't close this PR just yet as I will look into making the syntax type less confusing.

coastalwhite:
Having look at the draft instead of the live version, it is actually already changed. Therefore closing the PR.

SebastianZ:
Hah, Indeed! 😄 I just focused on the syntax and also didn't realize that this already got corrected in the draft.

Nonetheless, the grammar is still not clearly defined / incorrect. I'll check whether there's already a general discussion about the grammar used in the spec. and create an issue if needed.

Sebastian

PR #823, opened 2021-02-08
[color] Double link wraps in Bikeshed

nschonni:
The plain `<a>` wrapping was making a nested link with the `'` linking.

/cc @tabatkins in case there was a better wrapping to use here


tabatkins:
Nope, that's all correct.

PR #825, opened 2021-02-16
Use constructor operation for ShadowAnimation

foolip:
The Constructor extended attribute has been removed from Web IDL.


foolip:
@tabatkins can you review this one?

PR #844, opened 2021-04-22
Add default value {} for optional DOMMatrix2DInit arguments

foolip:
This is required by Web IDL, but because DOMMatrix2DInit is defined in
another spec wasn't noticed until all spec IDL was validated together.

See https://github.com/w3c/webidl2.js/issues/407 for a discussion about
the underlying validation problem.


foolip:
@tabatkins can you review? @dontcallmedom when this is merged into main, can you update the master branch to deploy?

PR #855, opened 2021-06-06
Add `interfaces` for <discard> definition

### LINKED ISSUE #854: [animations] `<discard>` lacks `DOM Interfaces:`
https://svgwg.org/specs/animations/#DiscardElement

image

It should be linked to SVGDiscardElement.

saschanaz:
Fixes #854

### DIFF
--- a/specs/animations/master/Overview.html
+++ b/specs/animations/master/Overview.html
@@ -24,7 +24,7 @@
     <dd><edit:previousversion/></dd>
     -->
     <dt>GitHub repository:</dt>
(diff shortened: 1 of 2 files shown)


birtles:
Thank you!

saschanaz:
@birtles The published date on the spec changed but not the actual content. Did something go wrong? 🤔

birtles:
Hmm, I have no idea how the build process works for SVG specs these days. I'm not even sure who does. @jwatt or @heycam: Do you know how the SVG Animations spec is generated these days? It looks like it only partially updated (perhaps something is looking at master instead of main?).

PR #878, opened 2022-03-19
Correct sweep-flag in circle/ellipse equivalent paths

### LINKED ISSUE #765: Incorrect sweep-flag in circle/ellipse equivalent path
The equivalent path of a circle (similar for ellipses) is currently defined as:

https://svgwg.org/svg2-draft/shapes.html#CircleElement

> The rx and ry parameters to the arc commands are both equal to the used value of the r property, after conversion to local user units, while the x-axis-rotation, the large-arc-flag, and the **sweep-flag are all set to zero**. The coordinates are computed as follows, where cx, cy, and r are the used values of the equivalent properties, converted to user units:

```
A move-to command to the point cx+r,cy;
arc to cx,cy+r;
arc to cx-r,cy;
arc to cx,cy-r;
arc with a segment-completing close path operation.
```

I think this is incorrect, that is, the sweep-flag should be 1, not 0, like it is for rounded rects for example (both are traced clockwise).

With the current equivalent paths, I get this for some of the examples in the spec, showing that:
- Path-data elliptical arcs are drawn correctly
- Rounded rects using 1 as sweep-flag are rendered correctly
- Circles and ellipses using 0 as sweep-flag are rendered incorrectly

image

AtkinsSJ:
Closes #765

### DIFF
--- a/master/shapes.html
+++ b/master/shapes.html
@@ -214,9 +214,9 @@ radius.</p>
   clock-wise direction (before any transformations).
   The <var>rx</var> and <var>ry</var> parameters to the arc commands
   are both equal to the used value of the <a>'r'</a> property, after conversion to local user units,
-  while the <var>x-axis-rotation</var>,
-  the <var>large-arc-flag</var>,
-  and the <var>sweep-flag</var> are all set to zero.
+  while the <var>x-axis-rotation</var>
+  and the <var>large-arc-flag</var> are set to zero,
+  and the <var>sweep-flag</var> is set to one.
   The coordinates are computed as follows,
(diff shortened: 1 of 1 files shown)


AtkinsSJ:
Hi, is there anything else I need to do for this? I did fill in the commitment form in March, but forgot to tell anyone afterwards that I'd completed it.

awesomekling:
cc @dirkschulze @AmeliaBR

AtkinsSJ:
@caribouW3 Sorry for the ping, but you seem the most active with merging PRs. Can you take a look please?

dalboris:
I am not in the committee but the author of the bug report (https://github.com/w3c/svgwg/issues/765), and I have reviewed this change and approve it. It really should be merged.

Issue #707, opened 2019-07-01
Referencing SVGs through USE from other domains


waterplea:
Currently, trying to host SVGs with symbols on CDN and then using them on your page through USE tag to allow color changes with `currentColor` results with error: `Unsafe attempt to load URL`. There is a bug for chromium suggesting it should follow CORS policies providing ability to allow such references:
https://bugs.chromium.org/p/chromium/issues/detail?id=470601

There was a suggestion to add crossorigin attribute to USE but it seems that specs only define it on IMAGE or SCRIPT:

https://svgwg.org/svg2-draft/linking.html#processingURL-fetch

Please have a look at it. Seems to me that it would be a lot safer to use SVGs like that rather than loading them with network request and inlining in the page which is a current workaround if you want to control the color.

AmeliaBR:
I agree that this is important.  It is a significant limitation for SVG the crossorigin `use` references aren't supported.  I think you make a good point, that this encourages *less* safe practices by authors.

When the `crossorigin` attribute was originally added to SVG, `<use>` was included. It was removed in 2015 because of concern about the details about how `<use>` imports should work.  But even then, the plan was to add it back in eventually. (See [discussion minutes & resolution](https://www.w3.org/2015/06/09-svg-minutes.html#action02))

Many of the issues that were brought up in that discussion (about how the use element shadow DOM model should work & how external files should be parsed) have since been resolved (although the current spec text doesn't reflect all the resolutions that have been made in other issues).  So it may be worth reconsidering.

waterplea:
I'm not a security expert, I tried USE with external SVGs on the same domain and it seems that scripts are not run, neither event handlers like onclick trigger and other basic malicious scripting I could think of. I realize there is more danger there that I do not know about, but like I said, people currently work around this by simply requesting SVG source and inlining it into the page and I doubt many of them run it through DOMPurify or something. So even if it might be a not that safe practice, to achieve the same effect people most likely turn to an even more dangerous solutions.

Glad to see it's not dismissed completely, hope you can think of something!

AmeliaBR:
@fsoder @longsonr @litherum Consider this a request for feedback (please pass on to whichever colleague can best answer):

How easy/difficult would it be to make `<use>` cross-origin, with `crossorigin` controls? The default behavior (no attribute) would still be to block cross-origin requests. Processing of the external file would be the same as for same-domain external `<use>` references.

longsonr:
[...]

PR #901, opened 2022-12-09
Remove support for data: URL in SVGUseElement

shhnjk:
### Motivation
Assigning an attacker controlled string to `SVGUseElement.href` causes XSS due to data: URLs. This also led to a [bypass of Trusted Types](https://github.com/w3c/trusted-types/issues/357) in Blink.

Additionally, data: URLs can only trigger script execution in script loaders such as HTMLScriptElement.src or [dynamic import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import). However, SVGUseElement is an exception to this, which also caused a [bypass](https://bugs.chromium.org/p/chromium/issues/detail?id=1306450#c10) in the Sanitizer API. We believe that this also led to several other bugs in sanitizers and linters missing a check for this special case.

Since Webkit does not support data: URLs in SVGUseElement, Blink is planning to remove support for it. And therefore, we'd like to update the spec.

We have also [requested Mozilla's position](https://github.com/mozilla/standards-positions/issues/718) on this.

Currently, the [usage](https://chromestatus.com/metrics/feature/timeline/popularity/4356) of data: URLs in SVGUseElement is about 0.0056% of page load in Chrome.

### DIFF
--- a/master/struct.html
+++ b/master/struct.html
@@ -599,7 +599,7 @@ without having to ensure that it has an ID on its root element.</p>
 
 <p>
 User agents may restrict external resource documents for security
-reasons. In particular, this specification does not allow cross-origin resource requests in <a>'use'</a>.
+reasons. In particular, this specification does not allow cross-origin and <a href="linking.html#TermDataURL">data URL</a> resource requests in <a>'use'</a>.
 A future version of this or another specification may provide a method of securely enabling cross-origin re-use
 of assets.
 </p>
--- a/master/struct.html
+++ b/master/struct.html
@@ -600,8 +600,6 @@ without having to ensure that it has an ID on its root element.</p>
 <p>
 User agents may restrict external resource documents for security
 reasons. In particular, this specification does not allow cross-origin and <a href="linking.html#TermDataURL">data URL</a> resource requests in <a>'use'</a>.
-A future version of this or another specification may provide a method of securely enabling cross-origin re-use
-of assets.
 </p>
 
 <p>When the <a>'href'</a> attribute is set


shhnjk:
@caribouW3, could you PTAL? Thanks!

caribouW3:
Hi,
At a first glance, the proposal to forbid all data: url altogether seems a bit extreme. The current editors draft already clearly says that user agent may take that kind of restriction. If there's a global consensus that all user agent do the same thing, then it makes sense to reflect it in the specification.

mozfreddyb:
@shhnjk is there a WPT patch accompanying this standards change? It would be ideal if we had cases for cross/same-origin, -site, and same-document URLs too.

shhnjk:
> @shhnjk is there a WPT patch accompanying this standards change? It would be ideal if we had cases for cross/same-origin, -site, and same-document URLs too.

WIP WPT at https://github.com/web-platform-tests/wpt/pull/37511 🙂

shhnjk:
> I think we should essentially require a same origin URL here (and also enforce that for redirects, which would be easy if this was defined in terms of fetch...).
> 
> And no longer have "user agents may" or "a future version of". Let's actually settle this.

Thanks! Since https://github.com/w3c/svgwg/issues/707 got closed, I've removed the `a future version of` bit.
However, I think `User agents may restrict external resource documents for security reasons` bit is still necessary. For example, Blink enforced restrictions on external resource documents such as:

1. `<foreignObject>` is not supported.
2. Event handlers doesn't fire.

@caribouW3, I've got a [positive signal from Mozilla](https://github.com/mozilla/standards-positions/issues/718#issuecomment-1347915887) and [Webkit is likely supportive](https://github.com/WebKit/standards-positions/issues/108#issuecomment-1354354703) (pending formal decision until January 6th).
Is there anything else I need consensus on to land this change?

SebastianZ:
Just to provide a use case for cross-origin requests: I recently had the need to re-use an external SVG multiple times within a document and style them individually.

This only works via <use>. Though all images on that site are normally coming from a CDN. And because cross-origin requests are currently not allowed, I had to deliver them via the same domain instead.

If SVG obeyed to CORS, this wouldn't be an issue.

And there could also be some CORS header that controls whether data: URLs are allowed in SVG. So, restricting them by default, yes, but authors would still have a way to opt-in to support them.

Sebastian

shhnjk:
@SebastianZ, that was discussed in https://github.com/w3c/svgwg/issues/707#issuecomment-1356259989, and there are other alternatives to do this such as [CSS Linked Parameters](https://drafts.csswg.org/css-link-params/). And you can also fetch cross-origin SVG and get response as Blob, and then create a Blob URL to use it as a resource in SVGUseElement too.

@caribouW3, friendly ping for https://github.com/w3c/svgwg/pull/901#issuecomment-1358195469 🙂

caribouW3:
Since it will be supported in [webkit](https://github.com/WebKit/standards-positions/issues/108#issuecomment-1354354703), let's merge this.
@shhnjk, you can then merge the corresponding wpt PR.

shhnjk:
> Since it will be supported in [webkit](https://github.com/WebKit/standards-positions/issues/108#issuecomment-1354354703), let's merge this. @shhnjk, you can then merge the corresponding wpt PR.

Thank you! The WPT PR should be merged shortly (by the bot)!

mimer123 (approved):

SebastianZ:
> @SebastianZ, that was discussed in [#707 (comment)](https://github.com/w3c/svgwg/issues/707#issuecomment-1356259989), and there are other alternatives to do this such as [CSS Linked Parameters](https://drafts.csswg.org/css-link-params/).

Yes, CSS Linked Parameters will eventually solve the use case of applying different styles to a single external SVG resource.
And thanks for pointing me to the other issue! (I previously already commented there, as well, but totally forgot about it.)

Sebastian

donatocayurin:
Thank you ❤️🙏

donatocayurin:
❤️🙏

PR #924, opened 2023-10-21
Fix getElementByID IDL to allow null

### LINKED ISSUE #923: The IDL of SVGSVGElement.getElementById does not allow null
Per https://svgwg.org/svg-next/struct.html#__svg__SVGSVGElement__getElementById:

>The getElementById method, must return the first element in [tree order](https://dom.spec.whatwg.org/#concept-tree-order), within the ‘[svg](https://svgwg.org/svg-next/struct.html#SVGElement)’ element's descendants, whose ID is elementId, or null if there is no such element.

But the corresponding IDL is `Element getElementById(DOMString elementId);` which does not allow null.

saschanaz:
Fixes #923.


saschanaz:
@caribouW3 I just saw the last merge in January, can you also merge this? 👀

Ahmad-S792:
@sideshowbarker - I think we can merge `getElementsById-in-disconnected-svg.html` test case from WebKit repo to WPT.

sideshowbarker:
> @sideshowbarker - I think we can merge `getElementsById-in-disconnected-svg.html` test case from WebKit repo to WPT.

OK — Has somebody opened a PR for that?

PR #934, opened 2024-03-20
Upstream Trusted Types integration with SVGAnimatedString

lukewarlow:
This PR upstreams the trusted types integration with SVGAnimatedString. See https://www.w3.org/TR/trusted-types/#integration-with-svg

### DIFF
--- a/master/types.html
+++ b/master/types.html
@@ -2464,7 +2464,7 @@ be defined to additionally reflect a second, deprecated attribute.</p>
 
 <pre class="idl">[Exposed=Window]
 interface <b>SVGAnimatedString</b> {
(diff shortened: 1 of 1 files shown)


caribouW3:
It would be good to have a publish version of the trusted types spec instead of an editor's draft.

caribouW3 (changes requested):
Please refer to a /TR version rather than an editor's  draft.

caribouW3 (approved):

caribouW3 (approved):

lukewarlow:
@caribouW3 this doesn't appear to be updated in https://svgwg.org/svg2-draft/types.html#InterfaceSVGAnimatedString

Does this just need publishing or have I missed something in this PR that I need to change?

PR #935, opened 2024-04-02
Update painting.html stroke-width

JimJJewett:
(1)  Specify that the stroke-width is centered on the path, rather than inside or outside the shape. (2)  Correct(?) the stroke-width of a dash.

(A)  It is expected that the location of the stroke's own paint will be controllable in the next SVG edition, but the default/initial position will still need to be defined.

(B)  The algorithm for painting a dash collects all points on a line perpendicular to the path that are within a distance of stroke-width.  Because the path has two sides, this effectively doubles the stroke-width.  Either that should be corrected (as I proposed here) or it should be made explicit that dashes are double-thick.

### DIFF
--- a/master/painting.html
+++ b/master/painting.html
@@ -762,7 +762,8 @@ property</h3>
 
 <p>This property specifies the width of the stroke on the current object.
 A zero value causes no stroke to be painted. A negative value
-is <a>invalid</a>. A <a>&lt;number&gt;</a> value represents a value in <a>user units</a>.</p>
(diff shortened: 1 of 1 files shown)


JimJJewett:
I consider this an editorial clarification, which I *think* is classified as non-substantive.  It isn't clear whether I can (or even should be able to) make that classification.

Alternatively, I'm probably willing to sign the appropriate waivers, though I didn't see a link to what they are.  (Apologies if they are visible to someone less oblivious and/or not using a high-contrast viewer.)

PR #936, opened 2024-04-02
Update paths.html H and V not equiv to 0

JimJJewett:
The existing text claimed that line commands H and h (and V and v) were equivalent; the capitalized versions should still be in absolute terms.  If, for backwards compatibility reasons, the existing text is correct, this should be called out explicitly.


JimJJewett:
I consider this an editorial clarification, which I think is classified as non-substantive. It isn't clear whether I can (or even should be able to) make that classification.

Alternatively, I'm probably willing to sign the appropriate waivers, though I didn't see a link to what they are. (Apologies if they are visible to someone less oblivious and/or not using a high-contrast viewer.)

PR #949, opened 2024-07-20
Correct link for SVGAElement.relList

jamierocks:
This corrects the link from .rel to .relList.


jamierocks:
As I understand, given this is editorial, I don't need to sign up to anything. Please let me know if that's wrong :)

caribouW3 (approved):

PR #999, opened 2025-08-25
Adding resolution of content dependent units used in sizing properties for inner SVG elements

goldenboy777:
This PR adds the CSS WG Resolution in [w3c/csswg-drafts#12376 (comment)](https://github.com/w3c/csswg-drafts/issues/12376#issuecomment-3210998345) for SVG.
 
> The CSS Working Group just discussed `` [css-sizing-3][css-values-4] Define `width` and `height` CSS values for SVG Elements in a mapping ``, and agreed to the following:
> 
> * `RESOLVED: content dependent units used in width and height for inner SVG elements resolve to SVG's definition of auto`

### DIFF
--- a/master/geometry.html
+++ b/master/geometry.html
@@ -399,7 +399,7 @@ and <a>'height'</a> on the <a>'svg'</a> element is treated as 100%.</p>
 <p>The value <span class='prop-value'>auto</span> for <a>'width'</a>
 and <a>'height'</a> on the <a>'image'</a> element is calculated from the referenced image's intrinsic dimensions and aspect ratio, according to the CSS <a>Default Sizing Algorithm</a>.</p>
 
-<p>Content dependent units used in 'width' and 'height' for inner SVG elements resolve to SVG's definition of <span class='prop-value'>auto</span></a>. These content dependent units include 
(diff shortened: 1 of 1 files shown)


goldenboy777:
@longsonr @shallawa @nikolaszimmermann, can you guys review this PR if possible?

@caribouW3 @sideshowbarker can any of you help in adding reviewers or merging this PR?

goldenboy777:
> Hi, It does not look like spec prose. What should go into the spec is only the sentence in the Resolution part. The other info should go into the original issue in GH. I wonder if it could also include an exhaustive list of the content-dependent units.

@caribouW3, I added the sentence as you suggested. Let me know if more work needed to be done.

content-dependent keywords was a term that I coined to make the discussion easier, there is no CSS spec which mention that as is, but this spec https://drafts.csswg.org/css-sizing-3/#propdef-width list the properties which `width` CSS property can take among them are the listed `min-content`, `max-content`, `fit-content` and `calc-size` which are dependent on CSS box model.
This in my opinion are the units where ambiguity was for SVG elements, so I have listed them, let me know your thought here.

PR #1003, opened 2025-09-03
Replace spec prose with IDL extended attribute for simple reflection

### LINKED ISSUE #991: Use new [Reflect] IDL extended attribute for SVG element properties that need to reflect.
See https://github.com/whatwg/html/pull/11450 - this can be used on stuff such as SVGStyleElement.

lukewarlow:
This aligns with the HTML spec in using [Reflect] rather than spec prose for the simple HTML-like attribute reflection cases.

Closes #991


lukewarlow:
cc @karlcow @sideshowbarker

Issue #1029, opened 2025-11-10
Specifying processing model and algorithm for `ping` and `referrerPolicy` in SVG `a` element


goldenboy777:
Chromium is planning to support the `hrefLang`, `type`, `ping`, and `referrerPolicy` attributes for SVGAElement ([Intent to Ship](https://chromestatus.com/feature/5140707648077824?gate=6224522713890816)), in alignment with the [SVG2 specification](https://svgwg.org/svg2-draft/linking.html#InterfaceSVGAElement).

Following the links related to referrerPolicy in the spec ([reference](https://svgwg.org/svg2-draft/linking.html#AElementReferrerpolicyAttribute)), it states:

>These attributes further describe the targetted resource and its relationship to the current document. Allowed values and meaning are [as defined for the a element in HTML](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element).

While the expected behaviour of these attributes should mirror that of HTML `a` elements, there is ambiguity regarding the processing model and algorithms, particularly for `ping` and `referrerPolicy`. 
This concern has been raised by Chromium API owners and echoed by WebKit contributors ([signal discussion](https://github.com/WebKit/standards-positions/issues/566)).

Should the SVG2.0, be updated to include this information explicitly? Should relevant links corresponding to processing model and algorithm of HTMLAnchorElement be given for that?

karlcow:
We should double check and probably create tests to understand what each browser is doing in SVG2 and if it aligns with what the [HTML processing model for `a` is doing](https://html.spec.whatwg.org/multipage/links.html#following-hyperlinks-2).

If the browsers are doing the same thing, then we could just point to HTML for the processing model. 
If not AND browsers are aligned for what they are doing in SVG, we need to describe the differences.

goldenboy777:
As per the ACTION on me I created one HTML spec issue https://github.com/whatwg/html/issues/11936 asking whether we can include SVG `a` element in the HTML spec as well.

@karlcow who should I tag in the issue so that the HTML experts can review it?

karlcow:
@goldenboy777 I see the discussion well going on on the htmlwg.

PR #1037, opened 2025-11-25
Adding the note for SVGAElement regarding processing model and algorithms

goldenboy777:
As per the discussion in [Specifying processing model and algorithm for ping and referrerPolicy in SVG a element](https://github.com/w3c/svgwg/issues/1029), SVGAElement should also follow the processing model and algorithm as defined for HTMLAElement https://html.spec.whatwg.org/multipage/links.html#following-hyperlinks-2.

However, the corresponding change should ideally be done/referred in the HTML spec, as per the resolution in [Specifying processing model and algorithm for ping and referrerPolicy in SVG a element](https://github.com/w3c/svgwg/issues/1029):
`RESOLUTION: let's create a note in the SVG spec mentioning that we are waiting on the HTML spec to clarify the processing model for these attributes.`

This PR tracks the work done for adding the above note


karlcow (approved):
This looks good to me.

PR #1059, opened 2026-02-08
Clarify that CSS width/height properties only apply to outer svg elements

goldenboy777:
Per the [SVG WG resolution from 2026-02-05](https://www.w3.org/2026/02/05-svg-minutes.html#cba7):

> The CSS width/height properties are honored as presentation attributes for rect, image, foreignObject and outer-SVG elements and ignored by all other elements with the width/height attributes.

This change updates Section 7.8 (Sizing properties) to:

- Clarify that CSS `width` and `height` properties apply only to the **outermost** `<svg>` element, not nested ones
- Add a note explicitly stating that nested `<svg>`, `<use>`, and `<symbol>` elements have width/height attributes but do **not** accept them as CSS properties—they must be specified as XML attributes

### DIFF
--- a/master/geometry.html
+++ b/master/geometry.html
@@ -401,22 +401,41 @@ may be constrained by the value of the
 <a href="https://www.w3.org/TR/CSS21/visudet.html#propdef-max-height"><span class="prop-name">min-height</span></a> properties.
 </p>
 
-<p>The value <span class='prop-value'>auto</span> for <a>'width'</a>
-and <a>'height'</a> on the <a>'svg'</a> element is treated as 100%.</p>
+<p>The following table summarizes how the value
+<span class='prop-value'>auto</span> for <a>'width'</a> and <a>'height'</a>
+is treated on different elements:</p>
 
-<p>The value <span class='prop-value'>auto</span> for <a>'width'</a>
-and <a>'height'</a> on the <a>'image'</a> element is calculated from the referenced image's intrinsic dimensions and aspect ratio, according to the CSS <a>Default Sizing Algorithm</a>.</p>
-
-<p>Content dependent units used in 'width' and 'height' for inner SVG elements resolve to SVG's definition of <a><span class='prop-value'>auto</span></a>. These content dependent units include 
-<a href="https://drafts.csswg.org/css-sizing-3/#valdef-width-min-content">'min-content'</a>, <a href="https://drafts.csswg.org/css-sizing-3/#valdef-width-max-content">'max-content'</a>, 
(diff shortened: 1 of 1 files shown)


goldenboy777:
@dirkschulze @karlcow @nikolaszimmermann, I have raised this pull request as per the resolution on [SVG WG resolution from 2026-02-05](https://www.w3.org/2026/02/05-svg-minutes.html#cba7), let me know if it's okay to merge this PR

master/geometry.html
```diff
@@ -377,12 +377,20 @@ <h2 id="Sizing">Sizing properties: the effect of the
...
 The properties are also used for laying out embedded elements from the HTML namespace.
 </p>
 
+<p class='note'>
+Other SVG elements, such as nested <a>'svg'</a> elements, the <a>'use'</a> element, and
+the <a>'symbol'</a> element, also have <span class='attr-name'>'width'</span> and
+<span class='attr-name'>'height'</span> attributes, but do not accept them as CSS properties.
+On these elements, the width and height must be specified as XML attributes; CSS properties
+with the same names are ignored.
+</p>
```
karlcow: I wonder if there should be markup examples of what is working and not working. 
Also probably all of  these requires WPT tests.

karlcow (approved):
Thanks for editing. Just a nit comment about potential examples.

master/geometry.html
goldenboy777: We already have few WPTs for nested/inner svg already, (which I need to update as per the resolution)
https://wpt.fyi/results/svg/styling?label=master&label=experimental&aligned&q=nested

But I can add some more for the all the elements we are targeting.

master/geometry.html
```diff
@@ -377,12 +377,20 @@ <h2 id="Sizing">Sizing properties: the effect of the
 
 <p>The CSS <a>'width'</a> and <a>'height'</a> properties are used for
 sizing some SVG elements.  Specifically, they are used to size
-<a>'rect'</a>, <a>'svg'</a>, <a>'image'</a> and
-<a>'foreignObject'</a>.  All of these elements have <span class='attr-name'>'width'</span>
+<a>'rect'</a>, <a>'image'</a>, <a>'foreignObject'</a>, and
+the outermost <a>'svg'</a> element.  All of these elements have <span class='attr-name'>'width'</span>
```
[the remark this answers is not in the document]
goldenboy777: Hi @karlcow, did you get the chance to review the new changes? wanted someone to re-look at the changes once before I merge

Issue #828, opened 2021-03-01
UA stylesheet seems to require title as display:  none but browsers do display: inline


domenic:
https://svgwg.org/svg2-draft/styling.html#UAStyleSheet says

```css
defs,
clipPath, mask, marker,
desc, title, metadata,
pattern, linearGradient, radialGradient,
script, style,
symbol {
  display: none !important;
}
```

(in the SVG namespace). However, browsers do not appear to respect this: http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8961

At least one web developer seems to expect inline, not none: https://github.com/jsdom/jsdom/issues/3064

It's possible I'm reading the spec wrong?

heycam:
I think the intention of that rule in UA style sheet is to formalize the fact that these elements never render, and to make it clear that e.g. you can't add `style="display: inline"` to a `defs` and make it render.  Despite this, there is the definition of [never-rendered elements](https://svgwg.org/svg2-draft/render.html#TermNeverRenderedElement), which should also prevent them from rendering.

There's precedent in the HTML spec for [having UA style sheet rules](https://html.spec.whatwg.org/#hidden-elements) make certain elements be display:none, though it looks like most of them aren't also !important, and would actually render if made display:inline.

domenic:
Thanks, that makes sense. Do you know if any browsers in the working group plan to actually implement this UA stylesheet, or would it be best to remove it and just use never-rendered?

heycam:
Since nobody implements that right now, and it's not super important, we should probably just remove it.

karlcow:
There is a SVG css
https://searchfox.org/wubkat/source/Source/WebCore/css/svg.css
which is different than this one. Thanks @nikolaszimmermann

goldenboy777:
SVG css on the Chromium, which also doesnt contain the above style, anyway
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/css/svg.css?q=svg.css

PR #1071, opened 2026-03-05
#fix 828 - Drop display properties from UA svg CSS.

karlcow:
This is not currently used by any browsers and these elements are already marked in the spec with their specific behavior. 
https://github.com/w3c/svgwg/issues/828
https://www.w3.org/2026/03/05-svg-minutes.html#c364


goldenboy777 (approved):
The PR looks good to  me

qsadyysy786-tech (approved):

Issue #773, opened 2020-01-10
Support pathLength via CSS


argyleink:
I'd expected to be able to set this attribute via CSS but turns out that I can't. It was also pointed out here in this tutorial on an svg animation by @shshaw https://youtu.be/7aobRPg7BXI?t=741

shshaw:
I would definitely love to see this supported. Being able to set pathLength with CSS would make stroke effects & animations even easier. I can imagine some great effects by animating pathLength as well: https://codepen.io/shshaw/pen/KKwRMZd?editors=1010

tabatkins:
Hm, I'm surprised `path-length` isn't a geometry property for it, yeah.

AmeliaBR:
I agree that it makes sense to treat `pathLength` as another geometry property. And I also agree that the logical property name would be `path-length` (so that the camel-cased DOM version is still `style.pathLength`.

So the question is: who's interested in implementing it? What spec version should this be targetted for?

fsoder:
Given sufficiently positive signals from other engine vendors, I'd be happy to implement (or aid an implementation of) the property (and presentation attribute mapping) in Blink.

grorg:
Consider this a positive signal from WebKit!

Tavmjong:
I've done some basic testing. The attribute 'pathLength' is supported by Firefox and Chrome, not by Safari and Edge. Inkscape does not support 'pathLength' but it shouldn't be hard to implement.

The attached zip file contains tests of stroke dashing using 'pathLength':
* path_length.svg: Tests basic support on path, rectangle, circle.
* path_length2.svg: Tests support on a very long spiral path consisting of cubic Beziers. Firefox and Chrome differ in the calculation of the path length by about 0.1% (14px out of 11760px). They both agree within 0.2% of my naive calculation of the path length.
* path_length3.svg: A SMIL animation on the spiral path, with and without using 'pathLength'. There is no indication of a problem with computation time.

I see no problems in adding 'path-length' as a CSS property.

[path_length.zip](https://github.com/w3c/svgwg/files/4108982/path_length.zip)

fsoder:
Chromium/Blink bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1045430

I looked at what tests existed for the 'pathLength':

https://wpt.fyi/results/svg?label=master&label=experimental&aligned&q=pathlength

That seemed to cover the visual tests, and there are a few related to SVGGeometryElement APIs in types/scripted/.

[...]

PR #1073, opened 2026-03-11
Add path-length CSS property definition (Section 9.6.2)

viralipurbey:
**Summary**

Add a CSS property definition for `path-length` alongside the existing `pathLength` SVG attribute in the Paths chapter (Section 9.6.1), per the [CSS WG resolution from 22 January 2020](https://lists.w3.org/Archives/Public/www-style/2020Feb/0009.html).

**Changes**

[paths.html](https://github.com/w3c/svgwg/blob/c6cc21c4561f626f8acc906c7ed53fc098724896/master/paths.html)

- Added new subsection 9.6.2 The 'path-length' property with:
     - CSS property definition table (`none | <number [0,∞]`>, initial value `none`, applies to shapes and `textPath`, not inherited, not animatable).
     - Prose describing the property's effect on 'distance-along-a-path' calculations ('text-on-path' layout, motion animation, stroke dash operations).
     - Statement that `pathLength` is a presentation attribute for `path-length`.
     - Note referencing the CSS WG resolution.
     
[definitions.xml](https://github.com/w3c/svgwg/blob/c6cc21c4561f626f8acc906c7ed53fc098724896/master/definitions.xml)

- Registered `path-length` in the presentation attributes category.
- Added `<property>` entry for `path-length`.
- Updated the `pathLength` attribute definition with `presentationattribute` mapping.

Addresses #773

**Preview**

image

### DIFF
--- a/master/changes.html
+++ b/master/changes.html
@@ -599,6 +599,14 @@ have been made.</p>
   </li>
 </ul>
 </div>
+<div class='changed-since-cr1 cr2 cr3'>
+<ul>
+  <li>Added the <a>'path-length'</a> CSS property
+    and made <a>'pathLength'</a> a presentation attribute for it.
+    <a href="https://github.com/w3c/svgwg/issues/773">Issue #773</a>
+    <a href="https://github.com/w3c/svgwg/pull/1073">Edits</a></li>
+</ul>
+</div>
 
 <h3 id="shapes">Basic Shapes chapter</h3>
 
--- a/master/paths.html
+++ b/master/paths.html
@@ -1243,6 +1243,87 @@ commands contribute to path length calculations.</p>
   </dd>
 </dl>
 
+<h3 id="PathLengthProperty">The <span class="property">'path-length'</span> property</h3>
+
+<table class="propdef def">
+  <tr>
+    <th>Name:</th>
+    <td><dfn id="PathLengthCSSProperty" data-dfn-type="property" data-export="">path-length</dfn></td>
+  </tr>
+  <tr>
+    <th>Value:</th>
+    <td>none | <a>&lt;number [0,∞]&gt;</a></td>
+  </tr>
+  <tr>
+    <th>Initial:</th>
+    <td>none</td>
+  </tr>
+  <tr>
+    <th>Applies to:</th>
+    <td><a>'path'</a>, <a>'circle'</a>, <a>'ellipse'</a>,
+    <a>'line'</a>, <a>'polygon'</a>, <a>'polyline'</a>, <a>'rect'</a></td>
+  </tr>
+  <tr>
+    <th>Inherited:</th>
+    <td>no</td>
+  </tr>
+  <tr>
+    <th>Percentages:</th>
+    <td>N/A</td>
+  </tr>
+  <tr>
+    <th>Media:</th>
+    <td>visual</td>
+  </tr>
+  <tr>
+    <th>Computed value:</th>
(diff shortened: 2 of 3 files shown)


karlcow (changes requested):
@viralipurbey Thanks for working on this. 
We should try to add the change to Changes section of the spec too. 
See an example in https://github.com/w3c/svgwg/pull/1071/changes#diff-c7a9bad640ef0fed758e524f8762e900e665c8005337bb9988a2dffc62d04d6dR516-R521

viralipurbey:
I have added it in changes section as well. Please review @karlcow @dirkschulze @goldenboy777

master/changes.html
```diff
@@ -599,6 +599,11 @@ <h3 id="paths">Paths chapter</h3>
   </li>
 </ul>
 </div>
+<ul>
+  <li>Added the <a>'path-length'</a> CSS property
+    and made <a>'pathLength'</a> a presentation attribute for it
+    (<a href="https://github.com/w3c/svgwg/issues/773">Issue #773</a>).</li>
```
goldenboy777: You have referenced the issue but not the PR itself. Other entries in changes.html consistently include both like above:
```
<a href="https://github.com/w3c/svgwg/issues/383">Issue discussion</a>
<a href="https://github.com/w3c/svgwg/pull/485">Edits</a>
```

Consider adding an `Edits` link to the PR:
```
<li>Added the <a>'path-length'</a> CSS property
  and made <a>'pathLength'</a> a presentation attribute for it
  (<a href="https://github.com/w3c/svgwg/issues/773">Issue #773</a>).
  (<a href="https://github.com/w3c/svgwg/pull/1073">Edits</a>)</li>
```

master/changes.html
```diff
@@ -599,6 +599,11 @@ <h3 id="paths">Paths chapter</h3>
   </li>
 </ul>
 </div>
+<ul>
```
goldenboy777, line 603: Do we need the `<div class=changed-since...>` here too?

master/paths.html
```diff
@@ -1243,6 +1243,87 @@ <h3 id="PathLengthAttribute">The <span class="attr-name">'pathLength'</span> att
...
+calibrate the user agent's own
+<a href="paths.html#DistanceAlongAPath">distance-along-a-path</a>
+calculations with that of the author.</p>
+
+<p>A value of <span class="prop-value">none</span> indicates that no
+author path length is specified and the user agent's own computed
+path length is used.</p>
+
+<p>A <a>&lt;number&gt;</a> value must be zero or greater.
+A value of zero is valid and must be treated as a scaling factor
```
goldenboy777, line 1299: The existing pathLength attribute description paths.html:1224-1227 includes important detail:

_A value of zero scaled infinitely must remain zero, while any non-percentage value greater than zero must become +Infinity._

The new CSS property prose (new line ~1299) only says:

_A value of zero is valid and must be treated as a scaling factor of infinity._

Should we add the clarifying sentence about zero-times-infinity behavior to keep the CSS property and attribute semantics consistent? or are these two separate things which I am confusing?

master/changes.html
viralipurbey: Done. Thanks for pointing out.

master/changes.html
viralipurbey: Added.

master/paths.html
viralipurbey: Added this for clarification. Thanks.

karlcow (approved):
Thanks a lot @viralipurbey 
This is wonderful.

PR #1077, opened 2026-03-19
Update <textPath> `path`/`href` fallback behavior per WG resolution

### LINKED ISSUE #393: Does the entire path need to be valid in order to use a path attribute on a textPath
Per https://www.w3.org/TR/SVG2/text.html#TextPathElement

If the ‘path’ attribute contains an error, the ‘href’ attribute must be used. 

I can understand that path="" or path="foo" might be ignored.

However, does this also mean that a path="M 100, 100 foo" is ignored? That seems inconsistent with path error handling where it says that they are rendered until an error occurs (https://www.w3.org/TR/SVG2/paths.html#PathDataErrorHandling)

I propose that path attributes be handled the same way as path rendering i.e. as long as the first element in the path is valid, the path would be used.

viralipurbey:
Clarifies the fallback behavior when the `path` attribute on a `textPath` element contains errors, per the [March 12, 2026 WG resolution](https://www.w3.org/2026/03/12-svg-minutes.html):

- If the `path` data results in no valid path segments (empty string, or error before any valid segments), fall back to `href`.
- If the `path` data produces at least one valid segment before the error, use the valid portion per [Error handling in path data](https://w3c.github.io/svgwg/svg2-draft/paths.html#PathDataErrorHandling); do not fall back to `href`.

Updated in both the TextPathElement precedence section and the `path` attribute definition.

Resolves [https://github.com/w3c/svgwg/issues/393](https://github.com/w3c/svgwg/issues/393)

### DIFF
--- a/master/text.html
+++ b/master/text.html
@@ -4094,8 +4094,17 @@
     <li><a>'xlink:href'</a> attribute</li>
   </ol>
   <p>
-    If the <a>'path'</a> attribute contains an error, the
-    <a>'href'</a> attribute must be used.
+    If the <a>'path'</a> attribute is specified but
+    the <a href="paths.html#PathDataBNF" class="syntax">path data</a>
+    results in no valid path segments (e.g., the value is an empty
+    string, or an error in the path data occurs before any
+    valid path segments are produced), then the
+    <a>'href'</a> attribute must be used instead.
+    If the <a>'path'</a> attribute contains valid path segments
+    before the first error, the path is used up to the error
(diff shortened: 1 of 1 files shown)


viralipurbey:
Can we please review this? @karlcow @dirkschulze @goldenboy777

goldenboy777 (commented):
Looks good to me, Thanks for making this change

nikolaszimmermann (approved):
LGTM

PR #1078, opened 2026-03-19
FIX #1018 getbbox returns 0,0,0,0 when values are invalid or absent.

### LINKED ISSUE #1018: Should getBBox returns values when the shape can't be drawn
Using this WPT test as an example
https://wpt.live/svg/types/scripted/SVGGraphicsElement.getBBox-03.html

```
  <rect id="rect1" x="1" y="2" width="-10" height="20"/>
```

The value for the [`rect`](https://svgwg.org/svg2-draft/shapes.html#RectElement) width is a negative number, which is invalid.

* CSS 2.1 Negative values for ['width'](https://www.w3.org/TR/CSS2/visudet.html#propdef-width) are illegal.

and in https://svgwg.org/svg2-draft/shapes.html#RectElement

> The [width](https://svgwg.org/svg2-draft/geometry.html#Sizing) and [height](https://svgwg.org/svg2-draft/geometry.html#Sizing) properties define the overall width and height of the rectangle :
1. **A negative value for either property is [invalid](https://svgwg.org/svg2-draft/types.html#TermInvalidValue) and must be [ignored](https://drafts.csswg.org/css-syntax/#css-ignored)**. 
2. A computed value of zero for either dimension **disables rendering of the element**.

Then in [4.4.2. Interface SVGGraphicsElement for getBBox](https://svgwg.org/svg2-draft/types.html#__svg__SVGGraphicsElement__getBBox)

> The getBBox method is used to compute the bounding box of the current element. When the getBBox(options) method is called, the [bounding box algorithm](https://svgwg.org/svg2-draft/coords.html#BoundingBoxes) is invoked for the current element, with fill, stroke, markers and clipped members of the options dictionary argument used to control which parts of the element are included in the bounding box, using the element's user coordinate system as the coordinate system to return the bounding box in. A newly created [DOMRect](https://www.w3.org/TR/geometry-1/#DOMRect) object that defines the computed bounding box is returned. **If [getBBox](https://svgwg.org/svg2-draft/types.html#__svg__SVGGraphicsElement__getBBox) gets called on a [non-rendered element](https://svgwg.org/svg2-draft/render.html#TermNonRenderedElement)**, and the UA is **not able to compute the geometry of the element**, then **throw an [InvalidStateError](https://heycam.github.io/webidl/#invalidstateerror)**.

Currently:

| Browser | x | y | width | height |
|---------|---|---|-------|--------|
| Firefox | 0 | 0 | 0     | 0      |
| Safari  | 0 | 0 | 0     | 0      |
| Chrome  | 1 | 2 | 0     | 20     |

So I decided to try with a `rect` were `width` would be missing.

```
  <rect id="rect1" x="1" y="2" height="20"/>
```

And still the results are the same. Chrome returns `0`, Firefox/Safari return 0 for everything as the shape is not valid. 

In [7.8. Sizing properties: the effect of the ‘width’ and ‘height’ properties](https://svgwg.org/svg2-draft/geometry.html#Sizing)

> The value auto for [width](https://svgwg.org/svg2-draft/geometry.html#Sizing) and [height](https://svgwg.org/svg2-draft/geometry.html#Sizing) on other elements is treated as 0.

but that's not the case here. We do not have a value `auto`.

The rules for parsing attribute values are in [4.2. Attribute syntax](https://svgwg.org/svg2-draft/types.html#syntax).

> The Initial value column gives the [initial value](https://svgwg.org/svg2-draft/types.html#TermInitialValue) for the attribute. When an attribute fails to parse according to the specified CSS Value Definition Syntax, ABNF or EBNF grammar, or if parsing according to the URL Standard or by the prose describing **how to parse the attribute indicates failure, the attribute is assumed to have been specified as the given [initial value](https://svgwg.org/svg2-draft/types.html#TermInitialValue)**.

There is no table for width in SVG2 as it refers to CSS.

The CSS 2.2 spec in 10.2 Content width: the ['width'](https://www.w3.org/TR/CSS22/visudet.html#propdef-width) property gives as initial value for width: `auto`

**but I'm still confused…**

It says ignore at the beginning, then `getBBOX()` can not be computed. 

Opinions?
* Should the WPT tests be modified and it should return `InvalidStateError`?
* OR Should the spec be modified to clarify what invalid width/height default to?

karlcow:


karlcow:
Hmm something wrong here. :) I will fix it.

nikolaszimmermann (approved):

PR #1079, opened 2026-03-19
Update from api.csswg.org/bikeshed to spec-generator

kfranqueiro:
This updates the Makefile to reference spec-generator instead of the discontinued CSSWG Bikeshed service.

I've verified that `make online` runs successfully with output similar to `make local`.


karlcow (approved):
Thanks a lot.

Issue #312, opened 2017-04-05
SVGAElement implements two conflicting interfaces


saschanaz:
The [SVGAElement](https://svgwg.org/svg2-draft/linking.html#InterfaceSVGAElement) in the SVG 2 Linking chapter implements two interfaces where both two define their own `href` attribute:

```WebIDL
SVGAElement implements SVGURIReference;
SVGAElement implements HTMLHyperlinkElementUtils;
```

```WebIDL
[NoInterfaceObject]
interface SVGURIReference {
  [SameObject] readonly attribute SVGAnimatedString href;
};
```

```WebIDL
[NoInterfaceObject]
interface HTMLHyperlinkElementUtils {
  stringifier attribute USVString href;
  readonly attribute USVString origin;
  attribute USVString protocol;
  attribute USVString username;
  attribute USVString password;
  attribute USVString host;
  attribute USVString hostname;
  attribute USVString port;
  attribute USVString pathname;
  attribute USVString search;
  attribute USVString hash;
};
```

The spec does not say which definition should be followed, and I think this conflict shouldn't in the spec. Is this a problem can be fixed?

AmeliaBR:
The SVG one should take precedence; it needs to, for backwards compatibility with SVG 1.1. 

But the IDL as written is definitely not clear about that.  I'm not sure if there's an easy way in IDL to represent that sort of polymorphic inheritance.  In other cases of a conflict (between an SVG 1.1 property and an interface with a same-named property), it's clear that the currently defined class member overrides the interface. But this is two different interfaces.

Either way, it's one more case of SVGAnimatedString being the bane of any attempts to harmonize SVG-DOM and HTML-DOM.

longsonr:
I suggest dropping HTMLHyperlinkElementUtils altogether and making the rest of the attributes the same as HTML as I don't see any use case in making them animatable (in fact download is specifically not animatable in the spec). i.e.

       [SetterThrows]
       attribute DOMString download;
       [SetterThrows]
       attribute DOMString rel;
       [PutForwards=value]
       readonly attribute DOMTokenList relList;
       [SetterThrows]
       attribute DOMString hreflang;
       [SetterThrows]
       attribute DOMString type;

dstorey:
As the IDL is written currently the SVG one doesn't take precedence. I believe in is invalid WebIDL. Edge for sure breaks when trying to mixin a member that already exists.

[...]

PR #1085, opened 2026-04-03
Fix CSS syntax of opacity-related properties

tidoust:
The spec used curly braces in the property values but only "single quotes" are allowed per the CSS syntax, see:
https://www.w3.org/TR/css-values/#component-types

This replaces curly braces with single quotes in the syntax definitions of `fill-opacity` and `stroke-opacity`.

### DIFF
--- a/master/painting.html
+++ b/master/painting.html
@@ -466,7 +466,7 @@ property</h3>
   </tr>
   <tr>
     <th>Value:</th>
-    <td>&lt;‘<a>'opacity'</a>’&gt;</td>
+    <td>&lt;'<a>'opacity'</a>'&gt;</td>
   </tr>
   <tr>
     <th>Initial:</th>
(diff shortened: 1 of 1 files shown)


karlcow (approved):
So weird. Thanks for spotting this. Maybe an auto-correct typo.

tidoust:
Thanks! FYI re. "spotting this", that's because we crawl specs to extract CSS syntaxes for publication in [Webref](https://github.com/w3c/webref) and in the [`@webref/css`](https://www.npmjs.com/package/@webref/css) mdn package in particular, which different CSS tools and IDEs rely on. We only publish valid syntaxes and have been patching these curly quotes in Webref for a few years already. In other words, the syntax problem was an old one, I just hadn't bothered reporting it yet because the spec wasn't being updated.

karlcow:
@tidoust if you identify other things, let me know. I'll fix them.

tidoust:
> @tidoust if you identify other things, let me know. I'll fix them.

Remaining ones for which we need to maintain patches seem to be known already:
- `SVGAElement` implementing conflicting interfaces, tracked in #312
- `SVGPathElement` being defined  both in SVG2 and SVG Paths (differently, and with incorrect IDL in SVG Paths). I see a [recent action plan](https://www.w3.org/2026/03/26-svg-minutes.html#4f1e) to re-introduce the path interfaces into SVG2, which seems good.

PR #1105, opened 2026-05-05
Exclude negative values using the CSS range notation

cdoublev:
This PR adds a [bracketed range](https://drafts.csswg.org/css-values-4/#css-bracketed-range-notation) to the CSS types that are used to define [`rx`](https://svgwg.org/svg2-draft/geometry.html#RxProperty), `ry`, [`stroke-miterlimit`](https://svgwg.org/svg2-draft/painting.html#StrokeMiterlimitProperty), [`stroke-width`](https://svgwg.org/svg2-draft/painting.html#StrokeWidthProperty), which do not allow negative values.

### DIFF
--- a/master/geometry.html
+++ b/master/geometry.html
@@ -122,7 +122,7 @@ of the position of the element. </p>
   </tr>
   <tr>
     <th>Value:</th>
-    <td><a>&lt;length-percentage&gt;</a></td>
+    <td><a>&lt;length-percentage [0,&infin;]&gt;</a></td>
   </tr>
   <tr>
     <th>Initial:</th>
--- a/master/painting.html
+++ b/master/painting.html
@@ -728,7 +728,7 @@ property</h3>
   </tr>
   <tr>
     <th>Value:</th>
-    <td><a>&lt;length-percentage&gt;</a> | <a>&lt;number&gt;</a></td>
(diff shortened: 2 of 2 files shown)


caribouW3:
Is the range restriction syntax applicable to length-percentage?

It is used in some CSS specs like https://www.w3.org/TR/css-align-3/, but it seems to me that the parsing model described in https://www.w3.org/TR/css-syntax-3/#ref-for-typedef-percentage-token②  does not work if there's no % sign, so the same interval can't constrain a length and a percentage.
CSS Values 4 does not say anything about applying range restriction to a percentage specifically, nor to length-percentage (or the other something-percentage shortcuts)

cdoublev:
Hmm, the comments in the discussions that preceded the generalization of this notation ([#7400](https://github.com/w3c/csswg-drafts/issues/7400), [#7584](https://github.com/w3c/csswg-drafts/issues/7584)) are not very much explicit than CSS V&U:

  > Values of `−∞` or `∞` must be written without units.

But I find 46 properties defined with `<length-percentage [0,∞]>`:

  - `border-*-clip`
  - `border-*-width`
  - `border-*-radius`
  - `border-limit`
  - `column-gap`
  - `font-size`
  - `height`
  - `line-height`
  - etc

Issue #974, opened 2025-04-05
New Interface SVGPathSegment: segments should be more editable


herrstrietzel:
When @longsonr informed me on StackOverflow that Firefox started to implement the pathData interface methods I stumpled upon the current not fully editable concept `SVGPathSegment` object. 

I wasn't sure if the current "read-only" state of [`SVGPathSegment`](https://svgwg.org/specs/paths/#InterfaceSVGPathSegment) was just due to the rather new implementation but it seems that this concept is actually intended by the specs as pointed out by 
**@lovasoa's issue report**. ("Clarification on [NoInterfaceObject] for SVGPathSegment and Usage of setPathData()")https://svgwg.org/specs/paths/#InterfaceSVGPathSegment

OK we can do some basic stuff as command coordinate rounding. But we can't create pathData from scratch.

I hope I'm just missing a point.   

If not: the whole idea of a standard/native browser parsing method is rather pointless if we can't fully manipulate the pathData including methods to:  
* add/remove Segments
* create pathData segments from scatch

While it's perfectly clear the native methods are not intended to replace a full-blown SVG-library we should be able to manipulate segments otherwise the concept of `setpathData()` is pointless – `getPathData()` would still be useful.  

In other words: maybe this is the right time to tweak the specs.

longsonr:
Adding and removing segments is something you can do already. It's just creating them that's hard.

Perhaps SVGPathSegment should not be nointerface object and should have a constructor instead.

jarek-foksa:
> Adding and removing segments is something you can do already. It's just creating them that's hard.

Correct me if I'm wrong, but the only way to create a segment on Firefox 137 is to use your own factory function that sets "d" attribute on a stub `<path>` and retrieves corresponding segment with `getPathData()`. 

> Perhaps SVGPathSegment should not be nointerface object and should have a constructor instead.

Or even better make `setPathData()` accept either `SVGPathSegment` instance or a segment-like plain object, in similar fashion to `DOMPoint.fromPoint()` which accepts `DOMPoint` or point-like object.

herrstrietzel:
I agree, ideally we should have constructors like `new PathSegment()`.  

> [...] in similar fashion to DOMPoint.fromPoint() which accepts DOMPoint or point-like object.

Not to forget the previously deprecated [`pathSegList` API model.](https://www.w3.org/TR/SVG11/paths.html#InterfaceSVGPathSeg).

From a – highly subjective – dev perspective (and also highly speculative):  
[...]

PR #1106, opened 2026-05-12
Change SVGPathSegment from [NoInterfaceObject] interface to dictionary

### LINKED ISSUE #1082: `SVGPathSegment` should be a dictionary, not a `[NoInterfaceObject]` interface
The SVG Paths spec ([§7.1](https://svgwg.org/specs/paths/#InterfaceSVGPathSegment)) currently defines `SVGPathSegment` as a `[NoInterfaceObject]` interface. This is problematic for three reasons:

1. `[NoInterfaceObject]` is deprecated in WebIDL. Modern WebIDL does not support this annotation; using it in a new spec is non-conforming.
2. Interface types prevent POJO compatibility. If `SVGPathSegment` is an interface, `setPathData()` requires actual `SVGPathSegment` wrapper objects rather than plain `{type, values}` JavaScript objects. This breaks the widely-used [path-data-polyfill](https://github.com/jarek-foksa/path-data-polyfill) (129+ stars), which creates plain objects. Firefox updated their code when it initially rejected POJOs and had to ship a separate fix in Firefox 138 ([bugzilla#1954044](https://bugzilla.mozilla.org/show_bug.cgi?id=1954044)).

Proposed change - replace the current WebIDL:

```
// Current:
[NoInterfaceObject]
interface SVGPathSegment {
  DOMString type;
  sequence<float> values;
};
```

with:

```
dictionary SVGPathSegment {
  required DOMString type;
  required sequence<unrestricted float> values;
};
```

This change bundles three improvements:

- `interface` → `dictionary`: Natively accepts POJOs, matching polyfill behavior and Firefox 138+.
- `float` → `unrestricted float`: SVG's error model is "render what you can." Path data can legally contain degenerate values (e.g., `A 0 0 0 0 0 0 0`). Using float would throw `TypeError` on NaN/Infinity, contradicting SVG's graceful-degradation model. `unrestricted float` matches Firefox's actual behavior and the polyfill.
- `required` keywords: Without required, `setPathData([{}])` would silently pass validation with undefined type and values. Both members must be present for a valid path segment.

All three changes are already aligned with Firefox's shipped behavior and the polyfill ecosystem. Chromium plans to implement with these semantics as well.

We already have one issue in editing WG: https://github.com/w3c/editing/issues/483

viralipurbey:
Resolves https://github.com/w3c/svgwg/issues/1082 and https://github.com/w3c/svgwg/issues/974.

Resolution: https://www.w3.org/2026/03/26-svg-minutes.html

- Convert SVGPathSegment from a [NoInterfaceObject] interface to a dictionary
- Mark type and values as required members
- Change values type from sequence<float> to sequence<unrestricted float>

### DIFF
--- a/specs/paths/master/Overview.html
+++ b/specs/paths/master/Overview.html
@@ -1363,26 +1363,25 @@ commands contribute to path length calculations.</p>
 
 <h2 id="DOMInterfaces">DOM interfaces</h2>
 
-<h3 id="InterfaceSVGPathSegment">Interface SVGPathSegment</h3>
+<h3 id="InterfaceSVGPathSegment">Dictionary SVGPathSegment</h3>
 
-The <a>SVGPathSegment</a> interface is a base interface that corresponds to a
+The <a>SVGPathSegment</a> dictionary corresponds to a
 single command within a path data specification.
 
-<pre class="idl">[NoInterfaceObject]
-interface <b>SVGPathSegment</b> {
-  DOMString type;
-  sequence&lt;float> values;
+<pre class="idl">dictionary <b>SVGPathSegment</b> {
+  required DOMString type;
+  required sequence&lt;unrestricted float> values;
 };</pre>
 
 <dl class="interface">
-  <dt class="attributes-header">Attributes:</dt>
+  <dt class="attributes-header">Members:</dt>
   <dd>
     <dl class="attributes">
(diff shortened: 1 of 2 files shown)


viralipurbey:
@dirkschulze @karlcow @goldenboy777 It's been some time we have raised this PR. Can we please review this?

karlcow (approved):
Thanks a lot @viralipurbey for working on this.

PR #1052, opened 2026-01-14
Swap SVGAElement to the new HyperlinkElementUtils

lukewarlow:
See https://github.com/whatwg/html/pull/12066 for context

Implementation Bugs:
- Chromium: https://issues.chromium.org/u/1/issues/512726601
- Firefox: ...
- WebKit: https://bugs.webkit.org/show_bug.cgi?id=314728

Tests:
- https://github.com/web-platform-tests/wpt/pull/59852
- https://github.com/web-platform-tests/wpt/commit/c2f97b77c80140864c172df829e50164c345e583

Fixes https://github.com/w3c/svgwg/issues/312

PR #1108, opened 2026-05-13
Remove hreflang and type IDL from SVGAElement

lukewarlow:
This is now on the HyperlinkElementUtils mixin

https://html.spec.whatwg.org/multipage/links.html#dom-a-hreflang


lukewarlow:
cc @sideshowbarker

this perhaps needs to wait till #1052 as currently this would just remove them from SVGAElement entirely which isn't what's desired.

lukewarlow:
cc @nikolaszimmermann here's a small follow up that would be good to merge too.

Issue #960, opened 2025-01-02
Clarify that SVG does not need to be valid XML


real-or-random:
The conformance section says:
"Additional conformance classes depend on whether the content is also valid and well-formed XML":

https://github.com/w3c/svgwg/blob/51ff4ba22b5d1d90f61fc6e388595d3e87d7839f/master/conform.html#L550-L551

I think it should say:
"Additional conformance classes depend on whether the content is also well-formed XML".

None of the conformance classes require that the document is *valid* XML.  (The only requirement that seems to be "imported" from XML is the validity of the id attribute.) 

Moreover, none of the examples in the entire spec contain a `<!DOCTYPE svg>` declaration which would be required to be valid XML 1.0. (See [XML 1.0 Section 2.8 "Prolog and Document Type Declaration"](https://www.w3.org/TR/xml/#sec-prolog-dtd), which says "The document type declaration MUST appear before the first [element](https://www.w3.org/TR/xml/#dt-element) in the document.")  

In general, I think it will be helpful to explain in more natural language that:
 - SVG does not need to be valid XML. (Since SVG is defined in terms of XML, readers could reasonably expect that valid XML is required)
 - A DOCTYPE declaration is not necessary nor recommended. (This was removed in https://github.com/w3c/svgwg/commit/96e51895fb4438ec3bea0965c90afc668aaed01c for some reason.)

vlakoff:
I agree with the original report. The current wording in the conformance section is misleading because it suggests that SVG content may need to be _valid XML_, while in practice and in the rest of the specification only _well‑formed XML_ is required.

### Why this matters

* SVG 2 does not define or rely on any DTD.
Therefore, XML _validity_ (in the XML 1.0 sense) is not a meaningful conformance criterion.
* Browsers do not perform DTD‑based validation when parsing SVG.
* None of the examples in the specification include a `<!DOCTYPE>` declaration, which would be required for XML validity.
* Likewise, the examples do not include an XML declaration (`<?xml version="1.0"?>`), and authors are not expected to use one.
Both constructs are unnecessary and often undesirable in real‑world SVG usage.

### What authors actually need

SVG content must be **well‑formed XML**, meaning:
* elements are properly nested,
* attributes are correctly quoted,
* there is a single root element,
* no invalid characters appear,
* etc.

But SVG content **does not need to be valid XML**, and the specification should not imply otherwise.

### Suggested clarification for the conformance section

Here are two possible formulations that would make the intent unambiguous:

[...]

Issue #1088, opened 2026-04-12
[SVG] standalone or not ?


letochagone:
Why is this example described as standalone, when in its code we see standalone=no ?
Is this an error in the specification (or am I the one who didn't understand...which is very likely :))

https://www.w3.org/TR/SVG2/single-page.html#chapter-struct

![Image](https://github.com/user-attachments/assets/e4e7f518-259b-4949-a0c0-c469b89f3198)

vlakoff:
This is a genuine naming collision between two completely unrelated concepts, which occurs in several code examples across the specification.

The XML declaration `standalone="no"` is an XML processing instruction that tells parsers whether the document may depend on an external DTD (Document Type Definition). It has nothing to do with whether the SVG file is structurally independent or embedded inline within a parent document.

Conversely, when the spec's prose describes a **"stand-alone"** document or file (formally defined in section 2.4.5), it is using SVG architectural terminology. It simply means the SVG exists as its own complete file, as opposed to an inline fragment embedded inside a parent HTML or XML document.

Both concepts are independently correct, but they use the exact same word with completely different meanings:

- **XML declaration:** `standalone="no"` → May reference an external DTD definition.
- **Spec prose:** `"stand-alone"` → A complete, independent `.svg` file.

----

### Proposed Solution: Clean up the code snippets

Rather than patching a single section or adding a global terminology note, the most definitive and modern fix would be to **remove the `standalone` attribute from all code snippets across the specification**.

This applies to both values: `standalone="no"` appears in several examples, and `standalone="yes"` in a few others. Neither serves any purpose in SVG 2: since the specification has dropped DTD support entirely, there is no external DTD to reference or to declare independence from. Both values are equally obsolete, and their inconsistent presence across examples only adds to the confusion.

While the `<?xml ?>` declaration itself can still have niche utility (e.g. to explicitly declare encoding), removing the `standalone` attribute specifically would completely eliminate this confusing naming collision, modernize the examples, and leave the canonical prose "stand-alone" untouched.

What does the Working Group think about stripping the `standalone` attribute from the specification examples?

vlakoff:
Following the discussion in [#1109 (review)](https://github.com/w3c/svgwg/pull/1109#pullrequestreview-4469297943), it may be worth going further than what I proposed above: rather than only removing the `standalone` attribute, the Working Group could consider removing the `<?xml …?>` declaration entirely from the examples.

[...]

PR #1109, opened 2026-05-15
Clarify that SVG requires well-formed XML, not valid XML

vlakoff:
Addresses #960. cc @real-or-random

### Drop "valid and" from the Document Conformance Classes intro

The intro sentence currently reads:

> Additional conformance classes depend on whether the content is also valid and well-formed XML.

None of the conformance classes defined in this section require XML *validity* in the XML 1.0 sense — that is, conformance to a DTD. Only *well-formedness* is required. Removing "valid and" brings the intro into agreement with the conformance classes that follow it.

### Add a clarifying note

Adds a note immediately after the intro paragraph making two things explicit:

- SVG 2 defines no DTD, and user agents do not perform DTD-based validation.
- A `<!DOCTYPE>` declaration and an XML declaration (`<?xml …?>`) are neither required nor recommended for SVG documents served as `image/svg+xml`.

Without this note, a reader could reasonably infer from the surrounding XML-related language that some form of XML validity is expected. The note removes that ambiguity without touching any normative text beyond the one-word fix above.

## Tests

No web-platform-tests PR is needed. This is a purely editorial change: no normative requirements are added or removed.

### DIFF
--- a/master/conform.html
+++ b/master/conform.html
@@ -548,7 +548,15 @@ would likely use static mode.
   conformance with this specification is defined by whether
   the content is or can generate a conforming DOM.
   Additional conformance classes depend on whether the content
-  is also valid and well-formed XML [<a href="refs.html#ref-xml">xml</a>].
+  is also well-formed XML [<a href="refs.html#ref-xml">xml</a>].
+</p>
+
+<p class="note">SVG content is not required to be <em>valid</em> XML in the XML 1.0
+  sense (i.e., conforming to a DTD). SVG 2 defines no DTD, and user agents do not
+  perform DTD-based validation. In particular, SVG documents do not need a
+  <code>&lt;!DOCTYPE&gt;</code> declaration or an XML declaration
+  (<code>&lt;?xml …?&gt;</code>); including either is neither necessary nor
+  recommended for SVG content served as <code>image/svg+xml</code>.
 </p>
 
 <h3 id="ConformingSVGDOMSubtrees">Conforming SVG DOM Subtrees</h3>
--- a/master/conform.html
+++ b/master/conform.html
@@ -556,7 +556,8 @@ would likely use static mode.
   perform DTD-based validation. In particular, SVG documents do not need a
   <code>&lt;!DOCTYPE&gt;</code> declaration or an XML declaration
   (<code>&lt;?xml …?&gt;</code>); including either is neither necessary nor
-  recommended for SVG content served as <code>image/svg+xml</code>.
+  recommended for SVG content, whether served as <code>image/svg+xml</code>
+  or embedded inline in HTML.
 </p>
 
 <h3 id="ConformingSVGDOMSubtrees">Conforming SVG DOM Subtrees</h3>
--- a/master/conform.html
+++ b/master/conform.html
@@ -547,7 +547,7 @@ would likely use static mode.
   For SVG content, therefore,
   conformance with this specification is defined by whether
   the content is or can generate a conforming DOM.
-  Additional conformance classes depend on whether the content
+  Some of the following conformance classes depend on whether the content
   is also well-formed XML [<a href="refs.html#ref-xml">xml</a>].
 </p>
 


vlakoff:
Could someone with access to the W3C Repository Manager please mark this PR as non-substantive?

This change is purely editorial: it clarifies existing intent without adding or removing any normative requirements. No conformance class is modified; the only changes are dropping the word "valid" from one sentence and adding an explanatory note.

master/conform.html
```diff
@@ -548,7 +548,15 @@ <h2 id="DocumentConformanceClasses">Document Conformance Classes </h2>
...
-  is also valid and well-formed XML [<a href="refs.html#ref-xml">xml</a>].
+  is also well-formed XML [<a href="refs.html#ref-xml">xml</a>].
+</p>
+
+<p class="note">SVG content is not required to be <em>valid</em> XML in the XML 1.0
+  sense (i.e., conforming to a DTD). SVG 2 defines no DTD, and user agents do not
+  perform DTD-based validation. In particular, SVG documents do not need a
+  <code>&lt;!DOCTYPE&gt;</code> declaration or an XML declaration
+  (<code>&lt;?xml …?&gt;</code>); including either is neither necessary nor
+  recommended for SVG content served as <code>image/svg+xml</code>.
```
karlcow: The full sentence is:

**Additional** conformance classes depend on whether the content is also valid and well-formed XML.

Then there is a series of conforming requirements:

* 2.4.1. Conforming SVG DOM Subtrees. **No mention of XML validity**
* 2.4.2. Conforming SVG Markup Fragments **No mention of XML validity**
* 2.4.3. Conforming XML-Compatible SVG Markup Fragments **No mention of XML validity** (just valid XML **IDs**)
* 2.4.4. Conforming XML-Compatible SVG DOM Subtrees **No mention of XML validity** (just valid XML **IDs**)
* 2.4.5. Conforming SVG Stand-Alone Files **No mention of XML validity**
* 2.4.6. Error processing

The ambiguity is brought by **additional**, I would say. Additional to the ones  already listed below. Or this specific list. 

2.4.3 https://w3c.github.io/svgwg/svg2-draft/conform.html#ConformingSVGXMLFragments
is requiring well-formedness only. 

I would drop additional, and replacing it by: `Some of the following conformance classes depend …` or something similar. 

👍

karlcow (approved):
Yes,  this is right. I still find the sentence a bit ambiguous.

karlcow:
@caribouW3 I don't think I can change the IPR status for this contribution.

master/conform.html
vlakoff: You're right about "Additional" — it suggests classes defined elsewhere, whereas the intent is to introduce exactly the classes listed right after. That's a genuine ambiguity.

There's also a related precision issue: it was implied that all those listed classes depend on well‑formedness, which isn't uniformly true:

* **2.4.1 (DOM Subtrees):** no requirement at all — well‑formedness is a markup concept and doesn't apply to DOM structures.
* **2.4.2 (Markup Fragments):** only implicitly and conditionally, when served as an XML MIME type, because the definition defers to "the appropriate parser for the document MIME type".
* **2.4.3–2.4.5:** required unconditionally.

"Some of the following conformance classes depend on…" fixes both issues. I've included that change in this PR.

vlakoff:
The clarifying note was restricting its guidance to SVG content served as `image/svg+xml`. But inline SVG embedded in HTML is equally common, and the same applies there: a `<!DOCTYPE>` declaration and an XML declaration are neither necessary nor recommended in that context either.

I've extended the note to cover both cases.

karlcow (approved):
Thanks for the contribution.

master/conform.html
```diff
@@ -547,8 +547,17 @@ <h2 id="DocumentConformanceClasses">Document Conformance Classes </h2>
...
+  is also well-formed XML [<a href="refs.html#ref-xml">xml</a>].
+</p>
+
+<p class="note">SVG content is not required to be <em>valid</em> XML in the XML 1.0
+  sense (i.e., conforming to a DTD). SVG 2 defines no DTD, and user agents do not
+  perform DTD-based validation. In particular, SVG documents do not need a
+  <code>&lt;!DOCTYPE&gt;</code> declaration or an XML declaration
+  (<code>&lt;?xml …?&gt;</code>); including either is neither necessary nor
+  recommended for SVG content, whether served as <code>image/svg+xml</code>
+  or embedded inline in HTML.
```
real-or-random, line 560: I think this AI slop is a real change because it says that an XML declaration is not recommended. Before this PR, the draft didn't state this.

This contradicts many code examples in the document that have an XML declaration.

Also, it's questionable whether XML 1.0 does not require an XML declaration. For example, when a file is not in utf-8 (and there's no external encoding information), a declaration is the only thing that will help XML processors to figure out the encoding. See Appendix F of XML 1.0; it's non-normative but it's there.

I think this PR simply shouldn't have touched this aspect. #960 was never about this.

karlcow: In the current version of the spec, many examples contains an `<?xml version="1.0" standalone=…` (xml declaration) even when not necessary for the purpose of the example. I counted 47 of them in total.

> neither necessary nor  recommended for SVG content

is indeed probably too strong. It escaped my review. That said for the common use cases of SVG, I'm not sure people are dealing with xml declaration these days. Always this tension in between the usage and the model. 

What would you recommend, propose as a prose? We can still change things. This is why this is still in development.

vlakoff: Thanks for the review, @real-or-random.

I maintain that an XML declaration is neither necessary nor recommended for SVG 2 content. Here is my reasoning:

**SVG 2 has moved away from a pure XML‑document model**

SVG 2 defines conformance in terms of the DOM, not in terms of an XML document instance. In that respect, the shift mirrors what happened with XHTML: the XML‑document framing gradually gave way to a DOM‑centric one, and the infrastructure that came with it — DTDs, XML declarations, formal XML validity — became irrelevant to conformance.

**The encoding argument has limited practical scope on the Web**

You are right that an XML declaration is the only in‑band way to declare a non‑UTF‑8 encoding when no external encoding information is available. However, browser behavior depends heavily on context:

* When SVG is served over HTTP as `image/svg+xml`, the `charset` parameter of the `Content-Type` header takes absolute precedence over the XML declaration, per [RFC 7303](https://datatracker.ietf.org/doc/html/rfc7303). The in‑document declaration is effectively irrelevant — and in fact, specifying a different encoding there could be actively misleading, since it would be silently ignored.
* For inline SVG in HTML, the HTML parser is used rather than an XML parser, so the XML declaration is ignored entirely. Here too, a non‑UTF‑8 `encoding` attribute would be misleading rather than helpful.
* For locally loaded SVG files (`file://`), where no HTTP headers are present, the XML declaration encoding *is* read by the browser's XML parser — but this is a marginal case, outside the primary Web context that SVG 2 targets.

In practice, UTF‑8 is both the Web standard and essentially universal. The XML declaration encoding attribute has a real but narrow use case — non‑UTF‑8 encoding with non‑HTTP delivery — which is far from the common scenario the note is addressing.

**Consistency with the spec's own examples**

I was already aware that 47 examples in the current draft include an XML declaration. That is precisely part of the problem the note aims to address: the examples have accumulated `<?xml …?>` lines as boilerplate, often paired with `standalone="no"` or `standalone="yes"`, without those attributes serving any purpose in SVG 2. This is also the subject of issue #1088, where I suggested removing the `standalone` attribute from examples — though I deliberately stopped short of proposing to remove the XML declaration itself, out of caution and to avoid scope creep.

The note does not say that an XML declaration is *harmful* or *forbidden*. It says it is neither necessary nor recommended. I think that is accurate: there is no SVG 2 conformance scenario in which it is required, and its presence in examples today is largely historical habit rather than intentional guidance.

If the Working Group considers the wording too strong, one option would be to soften it to acknowledge the edge case of non‑UTF‑8, non‑HTTP delivery. For instance:

> In particular, SVG documents do not need a `<!DOCTYPE>` declaration. An XML declaration (`<?xml …?>`) is similarly not required; authors working with non‑UTF‑8 encodings and non‑HTTP delivery may still find it useful, but it is unnecessary for the common case.

That said, I leave it to the Working Group to decide whether such a refinement is warranted.

real-or-random: > What would you recommend, propose as a prose?

I think what this PR gets fundamentally wrong is that it confuses doctype declarations (which are related to DTD-based validity) with XML declarations. These are entirely different beasts.

I simply propose to remove the entire `<p class="note">` paragraph. Changing "valid and well-formed XML" to well-formed XML" is enough to solve #960.

Then there are three follow-up issues that can be addressed separately:
  - Whether a doctype declaration should be recommended against. The answer is probably yes, because the purpose of such a declaration is to specify a DTD but SVG does not have one. But it's the job of the WG to form an opinion here. (Note that this was included in an [earlier working draft](https://www.w3.org/TR/2015/WD-SVG2-20150709/intro.html#Namespace).
  - Whether an XML declaration should be recommended against. This is subtler. The `standalone` declaration is useless (as pointed out by in #1088), but the `encoding` parameter can be useful. If you ask me, a useful recommendation is to omit the `standalone` declaration (because it's pointless and thus confusing). And then a note could be added that says that the entire XML declaration can be omitted if the file is in utf-8, and this may be wise. But again, this needs a decision.
  - Then the examples should be made consistent with the decisions.

karlcow:
@real-or-random Thanks Tim. Done.
https://github.com/w3c/svgwg/pull/1141/changes

karlcow:
@real-or-random could you open the followup issues? Thanks a lot.

PR #1118, opened 2026-06-05
 Change path-length property value type from number to length

viralipurbey:
Per the CSSWG resolution (resolution, 20 May 2026), change path-length value type from `<number [0,∞]>` to `<length [0,∞]>`, following the standard SVG attribute → CSS property pattern.

Also adds explicit note that unitless pathLength attribute values map to px.

Addresses w3c/csswg-drafts#13901


viralipurbey:
@dirkschulze @karlcow @goldenboy777 Please review.

karlcow (approved):
Thanks a lot @viralipurbey

PR #1117, opened 2026-06-04
updates SVG2 to comply to the process 2025 wording requirements

deniak:
That PR adapts the SVG2 spec to the process 2025 wording/markup requirements.
It also adds support for CR snapshot and drafts.

PR #1127, opened 2026-06-09
fix IDL exclusion

deniak:
#1117 introduced [an unintended side effect](https://github.com/w3c/svgwg/pull/1117#issuecomment-4661192696) where all the elements under the 'edit' namespace are being replaced and thus, overriding the expected behavior from `doCompleteIDL`.
That PR skips the replacement for `edit:excludefromidl`

/cc @tidoust


karlcow (approved):
thanks.

PR #1133, opened 2026-06-16
Define `getPathSegmentAtLength()` behavior for `NaN` distance

### LINKED ISSUE #1128: `getPathSegmentAtLength()`: Define behavior for `NaN` distance
The spec states that the distance parameter to `getPathSegmentAtLength()` "shall be clamped to [0, total-length]". However, `NaN` cannot be meaningfully clamped to a numeric range.

**Proposed resolution**: Return `null` when distance is `NaN`. This aligns with the graceful-degradation model used elsewhere in SVG (no exceptions for invalid numeric inputs) and is consistent with `unrestricted float` semantics. 

Firefox currently returns null for NaN.

For `±Infinity`, clamping to `[0, total-length]` works as expected (`-Infinity` → first segment, `+Infinity` → last segment). Firefox behaves the same way.

https://svgwg.org/svg2-draft/paths.html#DOMInterfaces

viralipurbey:
Clarify that `getPathSegmentAtLength()` returns null when the distance parameter is `NaN`, since `NaN` cannot be meaningfully clamped to a numeric range. This aligns with graceful-degradation model used elsewhere in SVG and matches Firefox's shipping behavior.

Fixes #1128

### DIFF
--- a/specs/paths/master/Overview.html
+++ b/specs/paths/master/Overview.html
@@ -1525,7 +1525,8 @@ The <a>SVGPathData</a> interface corresponds to the typed value of the <a>'d'</a
       <dd class="operation">
         <div>Returns the <a>SVGPathSegment</a> which is <var>distance</var>
         units along the path, utilizing the user agent's distance-along-a-path
-        algorithm. The <var>distance</var> shall be clamped to the range
+        algorithm. If <var>distance</var> is NaN, returns null.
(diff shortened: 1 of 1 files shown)


viralipurbey:

karlcow (approved):
Thanks ! 🥳

PR #1134, opened 2026-06-16
Define `getPathSegmentAtLength()` behavior at segment boundary

### LINKED ISSUE #1129: `getPathSegmentAtLength()`: Define which segment is returned at exact boundary
When the distance argument to `getPathSegmentAtLength()` falls exactly on the boundary between two segments, the spec does not define which segment should be returned (the earlier or later one).

**Proposed resolution**: Return the later segment. This is consistent with `getPointAtLength()`, which returns the point belonging to the later segment at a boundary. 

Firefox returns the later segment. Explicitly specifying this removes an interoperability ambiguity.

https://svgwg.org/svg2-draft/paths.html#DOMInterfaces

viralipurbey:
Clarify that when the distance falls exactly on the boundary between two segments, the later segment is returned. This is consistent with `getPointAtLength()` behavior and matches Firefox's implementation.

Fixes #1129


viralipurbey:

karlcow (approved):
Thanks.

PR #1135, opened 2026-06-16
Clarify that `getPathData()` returns the attribute base value

### LINKED ISSUE #1130: Clarify that `getPathData()` returns the attribute base value, not the CSS resolved/animated value
The `SVGPathData` mixin spec text says it corresponds to "the typed value of the d geometry property," which implies the CSS resolved value. The spec is also silent on whether animated values (via SMIL <animate> or Web Animations API) are reflected.

This is ambiguous in two ways:

 1. If CSS d property and the d attribute diverge (due to cascade), which does `getPathData()` return?
 2. During active animation, does it return the interpolated animated value or the base value?

**Proposed resolution**: Clarify that `getPathData()` returns the base value of the d attribute (not the CSS computed value and not the animated value). 

This matches `getAttribute("d")` semantics and Firefox's shipping behavior.
`setPathData()` should similarly be defined as equivalent to `setAttribute("d", ...)`.

viralipurbey:
Clarify that `getPathData()` operates on the base value of the `d` attribute, not the CSS resolved value or animated value. Similarly, `setPathData()` is equivalent to `setAttribute('d', ...)`. This matches `getAttribute('d')` semantics and Firefox's shipping behavior.

Fixes #1130

### DIFF
--- a/specs/paths/master/Overview.html
+++ b/specs/paths/master/Overview.html
@@ -1410,7 +1410,9 @@ interface mixin <b>SVGPathData</b> {
       <dt id="__svg__SVGPathData__getPathData" class="operation first-child">sequence&lt;<a>SVGPathSegment</a>> <b>getPathData</b>(optional <a>SVGPathDataSettings</a> <var>settings</var>)</dt>
         <dd class="operation">
         <div>Returns the sequence of path segments that corresponds to the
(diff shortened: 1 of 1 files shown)


viralipurbey:

karlcow (approved):
thanks.

PR #1141, opened 2026-06-23
Improve #960 pr with simpler phrasing.

karlcow:
* Drop the note paragraph
* Keep the simple sentence about the well-formed nature of the XML markup.


karlcow:
self-approving.

Issue #1057, opened 2026-02-04
Should `width` and `height` apply to nested `svg` elements as CSS properties


goldenboy777:
Chromium shipped support in M141 for applying `width` and `height` as presentation attributes on nested `svg` elements, aligning Chrome’s behavior with the SVG 2 [specification](https://svgwg.org/svg2-draft/geometry.html#Sizing).
Chromestatus entry: https://chromestatus.com/feature/5178789386256384.

As part of this change, CSS properties such as `width` and `height` now correctly affect nested `svg` elements, making the following two examples render identically:

```
With CSS Properties for nested <svg> element:
<svg width="100px" height="100px">
  <svg style="width:50px;height:50px;">
    <circle cx="50px" cy="50px" r="40px" fill="green" />
  </svg>
</svg>
```
```
Without CSS Properties for nested <svg> element:
<svg width="100px" height="100px">
  <svg width="50px" height="50px">
    <circle cx="50px" cy="50px" r="40px" fill="green" />
  </svg>
</svg>
```

However, shortly after launch, users began reporting breakages on their sites[ User reports of a svg size regression [449170647] - Chromium](https://issues.chromium.org/issues/449170647). Many of these pages were relying on incorrect or previously-ignored styling, and once the feature took effect, those styles suddenly applied and caused layout regressions. As such the feature was moved to **experimental status** in 143.

**From an inter op perspective, Firefox and Safari** don't support applying CSS properties `width` and `height` to nested `svg` already.
https://github.com/mozilla/standards-positions/issues/1243
https://github.com/WebKit/standards-positions/issues/509

Based on the breakage observed and the UseCounter data https://chromestatus.com/metrics/feature/timeline/popularity/5677, it's fair to say if width and height started applying on nested `svg` it can cause significant web compatibility issues (layout regressions to be precise). 
**In that case should we explicitly mention this behaviour in the spec or should an opt-in mechanism be adapted for this feature?**

goldenboy777:
It seems we earlier used to have the sentence in the spec but got removed for some reason.

"For backwards compatibility, when the ‘width’ or ‘height’ properties are specified as a presentation attributes, only the values that match the `length` production shall be mapped to CSS. Any other value must be treated as invalid."

karlcow:
https://codepen.io/webcompat/pen/PwzaaBO

<img width="3164" height="760" alt="Image" src="https://github.com/user-attachments/assets/43fcad20-3cb6-4c59-879f-3471047dfcce" />

[...]

PR #1160, opened 2026-08-10
Correct presentational attribute table for `height` and `width`

Calme1709:
99eb17c updated `height` and `width` to not be presentation attributes for `<symbol>`, `<use>`, and non-outermost `<svg>`s - this commit updates the presentational attribute table to reflect that change.

### DIFF
--- a/master/styling.html
+++ b/master/styling.html
@@ -350,7 +350,18 @@ the presentation attribute name is the same as the property name, in lower-case 
   </tr>
   <tr>
     <td>
-      <a>'height'</a>, <a>'width'</a>, <a>'x'</a>, <a>'y'</a>
+      <a>'height'</a>, <a>'width'</a>
+    </td>
+    <td>
+      <a>'foreignObject'</a>,
+      <a>'image'</a>,
+      <a>'rect'</a>, and
+      the outermost <a>'svg'</a>
+    </td>
+  </tr>
+  <tr>
+    <td>
+      <a>'x'</a>, <a>'y'</a>
     </td>
     <td>
       <a>'foreignObject'</a>,


master/styling.html
```diff
@@ -350,7 +350,18 @@ <h2 id="PresentationAttributes">Presentation attributes</h2>
...
   <tr>
     <td>
-      <a>'height'</a>, <a>'width'</a>, <a>'x'</a>, <a>'y'</a>
+      <a>'height'</a>, <a>'width'</a>
+    </td>
+    <td>
+      <a>'foreignObject'</a>,
+      <a>'image'</a>,
+      <a>'rect'</a>, and
+      the outermost <a>'svg'</a>
```
karlcow, line 359: nit: the outermost <a>'svg'</a> should use the defined, exported term `the <a>outermost svg element</a>`.

karlcow (commented):
You could open an issue.
This also requires a contributor agreement. 

Thanks for the catch.

karlcow:
There are more issues related to this change. that I discovered when digging in the support.

karlcow (changes requested):
+1. Checked the table against all three engines and this row is right: width and height map on foreignObject, image, rect and the outermost svg, and nowhere else.

As mentioned in a comment, nit: `the outermost <a>'svg'</a>`. 

There is a defined and exported term, `outermost svg element` in struct.html, linked throughout coords.html. `the <a>outermost svg element</a>` would point the cell at the definition. 

Same unlinked phrasing landed in geometry.html in 99eb17c, worth fixing in both.

Worth adding "Part of #1057" to the description. Not closing it.
The prose elsewhere still contradicts the table. I will open an issue about it.

One divergence to record: Chrome still maps width and height on symbol. Filing that against Chromium, nothing needed in this PR.

Leaving the x/y row alone. It has its own problem, separate issue.

Issue #1161, opened 2026-08-10
The open-vs-closed subpath stroking rule is only in an informative note, and Gecko doesn't follow it


karlcow:
The difference between a [closed subpath](https://w3c.github.io/svgwg/svg2-draft/paths.html#TermClosedSubpath) and an [open subpath](https://w3c.github.io/svgwg/svg2-draft/paths.html#TermOpenSubpath) that happens to end on its own start point decides whether the two ends get a join or two caps. The spec states it once, in the [`closepath` section](https://w3c.github.io/svgwg/svg2-draft/paths.html#PathDataClosePathCommand):

> A closed subpath differs in behavior from an open subpath whose final coordinate is the initial point of the subpath. The first and last path segments of an open subpath will not be joined, even when the final coordinate of the last path segment is the initial point of the subpath. This will result in the first and last path segments being capped using the current value of ['stroke-linecap'](https://w3c.github.io/svgwg/svg2-draft/painting.html#StrokeLinecapProperty) rather than joined using the current value of ['stroke-linejoin'](https://w3c.github.io/svgwg/svg2-draft/painting.html#StrokeLinejoinProperty).

That passage is `class="note ready-for-wider-review"` (informative). There is no normative statement of the rule anywhere in the spec, and one engine does not implement it.

### Code example

Three drawings of the same square. Same four sides each time; only the ending differs.

```html
<svg width="150" height="150" viewBox="0 0 150 150">
  <rect x="35" y="35" width="80" height="80" fill="none" stroke="black" stroke-width="20"/>
</svg>

<svg width="150" height="150" viewBox="0 0 150 150">
  <!-- closed -->
  <path d="M 35 35 H 115 V 115 H 35 V 35 Z" fill="none" stroke="black" stroke-width="20"/>
</svg>

<svg width="150" height="150" viewBox="0 0 150 150">
  <!-- open, but the last point is the first point -->
  <path d="M 35 35 H 115 V 115 H 35 V 35" fill="none" stroke="black" stroke-width="20"/>
</svg>
```

`stroke-linejoin` is `miter` and `stroke-linecap` is `butt`, both by default. Per the note, the third one should show two butt caps meeting at the top-left instead of a mitered corner, leaving a notch in the outer corner.

### Results

Tested 2026-08-10.

| | `<rect>` | closed path | open path |
|---|---|---|---|
| WebKit — Safari Technology Preview 249 | mitered corner | mitered corner | **notch at the top-left corner** |
[...]

Issue #753, opened 2019-11-20
Rect decomposition is invalid


tatarize:
https://www.w3.org/TR/SVG/shapes.html#RectElement

According to the spec, if rx, and ry are 0 the path is not closed.

1. perform an absolute moveto operation to location (x+rx,y);
2. perform an absolute horizontal lineto with parameter x+width-rx;
3. if both rx and ry are greater than zero, perform an absolute elliptical arc operation to coordinate (x+width,y+ry), where rx and ry are used as the equivalent parameters to the elliptical arc command, the x-axis-rotation and large-arc-flag are set to zero, the sweep-flag is set to one;
4. perform an absolute vertical lineto parameter y+height-ry;
5. if both rx and ry are greater than zero, perform an absolute elliptical arc operation to coordinate (x+width-rx,y+height), using the same parameters as previously;
6. perform an absolute horizontal lineto parameter x+rx;
7. if both rx and ry are greater than zero, perform an absolute elliptical arc operation to coordinate (x,y+height-ry), using the same parameters as previously;
8. perform an absolute vertical lineto parameter y+ry
9. if both rx and ry are greater than zero, perform an absolute elliptical arc operation with a segment-completing close path operation, using the same parameters as previously.

The issue is that 9 is using the segment-completing close path operation. But, if `rx == ry == 0` and 9 does not apply there is no instruction that the path is closed. When we are dealing with a rectangle where `rx==ry==0` the resulting instructions without the conditional arcs is:

1. perform an absolute moveto operation to location (x+rx,y);
2. perform an absolute horizontal lineto with parameter x+width-rx;
4. perform an absolute vertical lineto parameter y+height-ry;
6. perform an absolute horizontal lineto parameter x+rx;
8. perform an absolute vertical lineto parameter y+ry

This is an open path.

dalboris:
Yes, I think the spec should decompose 9. into:

9. If both rx and ry are greater than zero, perform an absolute elliptical arc operation to coordinate (x+rx, y), using the same parameters as previously;
10. Close the path

Especially since segment-completing close path operations don't seem to be sure to make it to SVG 2.0. That being said, I think any implementer can read between the lines and understand that obviously, the rect should be closed.

tatarize:
[...]

PR #1162, opened 2026-08-10
fix #893 - drop the remaining segment-completing references

### LINKED ISSUE #893: 'closepath' command not implemented
The spec for the `closepath` command was added 4 years ago and hasn't been implemented by any browser yet.
(https://svgwg.org/svg2-draft/paths.html#PathDataClosePathCommand). Currently it only exists in the spec, and as a failing wpt test.

karlcow:
The section went away in f579d453 but four links to the term stayed, and they render as "@@ unknown term" in the built spec. Three of them are in the equivalent-path algorithms for rect, circle and ellipse.

The final arc now ends at the initial point and an ordinary closepath closes the subpath, so every equivalent path can be written in the path data syntax. The closing line has zero length; a note in the Basic Shapes chapter says so, and says why that does not change fill or stroke.

The closepath is unconditional, so a rect with square corners is now closed too. Previously only the rounded-corner step closed the shape, which left a plain rect open and its first corner unjoined. See #753.

Spelling out the closepath rather than relying on the last command landing on the start point matters because those two are not the same thing, and that rule is currently only a note. See #1161.


karlcow:
@viralipurbey If you have time to review that would be great.

PR #1166, opened 2026-08-27
Editorial. fix #1165. Clarify size variable.

### LINKED ISSUE #1165: Clarify `size` in the spec for consistency `viewport size`, `font size`
`types.html` has two algorithms that introduce a name and then fill a different one.

[§ SVGLength `value`](https://w3c.github.io/svgwg/svg2-draft/types.html#__svg__SVGLength__value), step 3:

> Let *viewport size* be a basis to resolve percentages against, based on the SVGLength's associated element and directionality:
> - has no associated element → *size* is 100
> - has an associated element and horizontal directionality → *size* is the width of the associated element's SVG viewport
> - has an associated element and vertical directionality → *size* is the height of the associated element's SVG viewport
> - has an associated element and unspecified directionality → *size* is the length of the associated element's SVG viewport diagonal

All four branches assign *size*. Step 5 then reads *viewport size*, which nothing ever filled.

Step 4 has the same problem with *font size*, in one branch of two. The "has no associated element" branch says *font size*, the "has an associated element" branch says *size*. Both fill the same name, so the first one shows which is intended.

[`convertToSpecifiedUnits`](https://w3c.github.io/svgwg/svg2-draft/types.html#__svg__SVGLength__convertToSpecifiedUnits) repeats both switches verbatim, with the same slip.

`size` appears 10 times across the two algorithms and is never introduced anywhere in the spec.

Editorial only. All three engines implement the intended reading, including the three separate percentage bases: in a `200x100` viewport, `x="50%"` gives 100, `y="50%"` gives 50, and `r="50%"` gives 79.0569. Same for `convertToSpecifiedUnits`, where an absolute 50 converts back to 25%, 50% and 31.6228% respectively.

The fix is to replace *size* with *viewport size* or *font size* to match the enclosing "Let" line, in all 10 places. Present since 2f9cc83a (2015-08-19), the commit that added the algorithms, so there is no earlier spelling to preserve.

Happy to send a PR.

Related: #211 is about the *directionality* concept step 3 branches on, which is a real question. This one is just the typo.

karlcow:


karlcow:
This is just editorial and clarify the spec. It doesn't change its meaning. Self merging.

